PR #65, opened 2022-11-29
Some performance improvements about Faults

nicolas-clausolles:
Two main modifications:
- modify the return type of Faults accessors to return PolyLine2D by reference instead of by value
- implement bounding box checks before checking if segments intersect each other

**WARNING** I voluntarily let some commented sections in the code to justify the changes, they should be removed before integration


drenard77810:
This suggestion has finally been taken into account. The issue is closed.
Thanks

PR #89, opened 2023-11-15
Fix issue88

fabien-ors:
Copy _bipts NeighMoving member correctly. Beware that target checkers (elements of _bipts) are not duplicated (shared memory) because it is currently not "cloneable" (incompatibility with cross-language inheritance)


drenard77810:
Fabien has fixed things correctly.

drenard77810:
Bug is closed

drenard77810 (approved):
OK. Final comment for this PR

PR #95, opened 2023-11-21
Branche dr test

drenard77810:
This PR will allow viewing the helps for the main classes that have just been added


fabien-ors (approved):
Seems legit ;)

PR #104, opened 2023-12-05
Fix Db::CreateFillRandom function (rank variable must be 1-based)

fabien-ors:
The "rank" variable corresponds to the sample numbers in a Database (1-based). It is optional and it is just for decoration. All functions from the gstlearn library manipulate indices (0-based) (except for [] operator in R package that uses 1-based indices in order to make the gstlearn compatible with the R convention).
Except for the "rank" variable (added by `flag_add_rank` argument) and for displaying locator ranks (x1, x2, etc...), each time you encounter "rank" term (either in documentation or in function names), you must consider it as an **index** (0-based). ex: Db::addSelectionByRanks should be renamed Db::addSelectionByIndices. This big cleanup (rank => index) will be done later.


xfreulon:
I do not see the necessity to maintain two conventions 0-based and 1-based indices and the use of two denominations ranks and indices seems to me unnatural. I would keep only a single convention in gstlearn (0-based if necessary) and sticks to it every where in gstlearn.

fabien-ors:
The "rank" variable and the locator rank (x1, x2 etc...) is inherited from RGeostats and has been kept for retro compatibility. I agree that all this needs to be clarified / simplified. Let's talk about it later!
Meanwhile, I think that the PR could be merged because the bug in createFillRandom has been fixed :)

xfreulon:
In the comment of the modified code, I saw a index with the 0-based convention in Db.hh line 7. The comment should be modified.

xfreulon:
Oups. Not line 7 but line 72.

PR #118, opened 2024-01-27
Eigen dr

drenard77810:
This seems a good time to merge with "dev" as all tests are successful (for me) and the main Matrix Algebra is now carried by AMatrix class.
It simply remains to plug this class everywhere in the code. (next task)


fabien-ors (approved):
review done

PR #120, opened 2024-02-20
Eigen dr

drenard77810:
New version where references to csparse has been almost completely suppressed


NDesassis (approved):
It seems correct. Nice!

NDesassis:
Ok.

PR #125, opened 2024-02-28
Eigen dr

drenard77810:
The cleaning procedure has reached a milestone: old file vario.c has been completely incorporated within Vario class (and daughters) or Classical class.


fabien-ors (changes requested):
CHANGELOG.Geoslib should not be modified (it is provided only for historical reason)
Please, prefer updating the file CHANGES in the root folder (gstlearn version currently under construction is 1.1.0)

fabien-ors (approved):

PR #127, opened 2024-03-10
Eigen dr

drenard77810:
This new version is up-to-date with the latest requirements for SUPREMA.
It includes several optimization to speed up reading and writing of large 3D files.
It also correct the bug when drawing a grid using raster option (the offset of the figure was not correct when a rotation was involve).


fabien-ors (approved):
Ok.

PR #155, opened 2024-05-07
Spde variance

drenard77810:
The variance calculation has been added in PDE API. All tests are successfull.


fabien-ors (approved):
Nice job.
Just a little remark that doesn't prevent from merging : Reference times should be updated in bench_SPDE.

PR #165, opened 2024-05-17
Fix _USE_HDF5 build error

pierre-guillou:
This commit fixes a build error (extra parenthesis) when enabling the `USE_HDF5` CMake option.

Enjoy,
Pierre


fabien-ors (approved):
Indeed ! Thx

PR #168, opened 2024-05-22
Add GitHub Actions workflow to do static tests on C++ code

pierre-guillou:
This PR introduces a new GitHub Actions workflow, called "check_code", that performs several checks on gstlearn's C++ code. It leverages the "Clang Tools" to ensure good C++ code quality.

This workflow is heavily inspired by a similar one in the topology-tool-kit/ttk repository that I authored a few years ago (after fixing everything in the codebase).

The workflow consists in 3 jobs:
- `check-formatting` uses `clang-format` (https://clang.llvm.org/docs/ClangFormat.html) to force C++ code to be correctly formatted, plus looks for several small text formatting issues (non-ASCII characters, CRLF line endings...)
- `check-warnings` uses `clang-check` (https://clang.llvm.org/docs/ClangCheck.html) to parse C++ code quite fast in order to check that the library compiles without compilation errors or warnings (this not yet the case...)
- finally, `lint-code` performs extended tests on the C++ code, using `clang-tidy` (https://clang.llvm.org/extra/clang-tidy/) and the Clang Static Analyzer (https://clang.llvm.org/docs/ClangStaticAnalyzer.html) to detect potential bugs or non-efficient C++ constructs. It also checks that the Doxygen documentation is processed without warnings.

For now, a lot (not to say all of them) of these job are failing and there is some work to do in gstlearn's code to make them pass. The simplest being fixing the few Doxygen warnings.

This PR is mostly for setting a quality target and opening discussion.

Some tools need a configuration file to work properly. In particular, this PR needs:
- [ ] a top-level `.clang-format` file that describes the C++ formatting style (https://clang.llvm.org/docs/ClangFormatStyleOptions.html) otherwise gstlearn code will be checked against the default style,
- [ ] a top-level `.clang-tidy` file that lists the rules (https://clang.llvm.org/extra/clang-tidy/checks/list.html) we want to enable for gstlearn. I can provide one if needed.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/check_code.yml
+++ b/.github/workflows/check_code.yml
@@ -1,0 +1,167 @@
+name: check_code
+
+on:
+  # Permit calling trigger
+  workflow_call:
+  # Permit manual trigger
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Manual'
+        required: false
+        default: ''
+
+concurrency:
+  # allow a failing job inside a matrix to stop the other jobs
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CMAKE_EXPORT_COMPILE_COMMANDS: ON
+
+
+jobs:
+
+  # ----------------------- #
+  # Check source formatting #
+  # ----------------------- #
+  check-formatting:
+    runs-on: ubuntu-22.04
+    steps:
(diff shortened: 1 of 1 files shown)


fabien-ors (approved):
Tests are not currently OK.
But workflow is useful in dev

PR #169, opened 2024-05-22
Fix Clang warnings on Ubuntu

pierre-guillou:
This PR fixes all Clang warnings on Ubuntu, in 3 categories:

- the `override` keyword was missing on several virtual member function declaration. While having a similar effect to the `virtual` keyword, in C++>=11 they have a distinct meaning (c.f. https://en.cppreference.com/w/cpp/language/override) and Clang wants it enforced.
- the wrong `delete` operator was used in the hdf5 (the correct one was `delete[]`).
- in `WarningMacro.hpp`, the `__APPLE__` macro (only defined on macOS) used to detect Clang has been replaced with `__clang__`. On Ubuntu + Clang, `__APPLE__` was causing Clang to output `-Wunknown-warning-option` warnings because it did not understand `GCC diagnostic push` pragmas.

Enjoy,
Pierre

### DIFF
--- a/include/Anamorphosis/AnamDiscrete.hpp
+++ b/include/Anamorphosis/AnamDiscrete.hpp
@@ -31,7 +31,7 @@ public:
 
   /// AAnam interface
   bool hasGaussian() const override { return false; }
-  int getNClass() const { return _nCut + 1; }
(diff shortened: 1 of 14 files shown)


fabien-ors (approved):
Thx

PR #173, opened 2024-05-24
GitHub Actions: Use package managers for getting Boost & Eigen

pierre-guillou:
This PR makes use of platform-dependent package managers to handle the installation of the Boost and Eigen dependencies.

More specifically, `apt` is used on Ubuntu, `brew` on macOS and `vcpkg` on Windows. Note that while the first two provide binary packages, `vcpkg` seems to be building from source.

The custom actions `install-boost-unix`, `install-boost-windows`, `install-eigen-unix` and `install-eigen-windows` are now useless and can be removed in a further PR.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/nonreg-tests_macos-latest-one.yml
+++ b/.github/workflows/nonreg-tests_macos-latest-one.yml
@@ -19,10 +19,6 @@ env:
   NUMPY_VERSION : "1.23.5"
   R_VERSION : "4.3.2"
   SWIG_ROOT : ${{github.workspace}}/swig_420b
-  BOOST_ROOT: ${{github.workspace}}/boost
-  BOOST_VERSION: "1.72.0"
-  EIGEN_ROOT : ${{github.workspace}}/eigen
-  EIGEN_VERSION : "3.4.0"
   LLVM_ROOT : /opt/homebrew
 
 jobs:
@@ -33,7 +29,7 @@ jobs:
     - uses: actions/checkout@v3
     
     - name: Install dependencies
-      run: brew install llvm
+      run: brew install llvm boost eigen
       
     #- name: Setup Python Version
     #  uses: actions/setup-python@v4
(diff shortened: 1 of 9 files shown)


fabien-ors (commented):
Is it possible to keep the boost and eigen versions plz ? (for having stabe run)
Thx !

pierre-guillou:
You mean forcing a specific version of the Boost & Eigen packages ? I don't think this is possible with every package manager.

On Ubuntu, we only get what's packaged at a given time: Boost 1.74 on Ubuntu 22.04 (https://packages.ubuntu.com/jammy/libboost-dev) and Eigen 3.4.0 (https://packages.ubuntu.com/jammy/libeigen3-dev).

On macOS with `brew`, we can select either Boost 1.85 or 1.76 (https://formulae.brew.sh/formula/boost). Eigen is only at 3.4.0 (https://formulae.brew.sh/formula/eigen).

On Windows with `vcpkg`, we may be able to select the Boost version we want ()https://vcpkg.link/ports/boost/versions), the default being the latest one, 1.85. Eigen is similar (https://vcpkg.link/ports/eigen3/versions) and defaults to 3.4.0.

To sum up, Eigen is at 3.4.0 on all platforms, Boost is 1.85 (the latest) on `brew` and `vcpkg` and older on `apt`.

It might be better to track what's proposed by default by `apt` and `brew`, since that's what's users/developers tend to use, rather than downloading & building from source.

fabien-ors:
The idea was to prevent us from having failed builds due to new versions of eigen / boost. I agree that users tend to use the last versions. So it's OK for this PR. I just hope that we won't have to adapt our code each time boost or eigen API is changed.

fabien-ors (approved):

pierre-guillou:
It may be nice sometimes to have an early warning that there is an API break occurring in our dependencies through the CI…

Issue #148, opened 2024-04-24
Backup issue: In class VectorT: Use of soon to be deprecated/removed functions and operators


fabien-ors:
from @nicolas-clausolles:
 
- `std::shared_ptr<T>::unique` : deprecated in C++17, removed in C++20 ([cppref](https://en.cppreference.com/w/cpp/memory/shared_ptr/unique)), used [here](https://github.com/gstlearn/gstlearn/blob/96dd11e08dba46336c628f14106a8d6fca6bd3d5/include/Basic/VectorT.hpp#L244)

- `std::vector<T>` comparison operators (all except `==`) : removed in C++20 ([cppref](https://en.cppreference.com/w/cpp/container/vector/operator_cmp)), used [here](https://github.com/gstlearn/gstlearn/blob/96dd11e08dba46336c628f14106a8d6fca6bd3d5/include/Basic/VectorT.hpp#L59). Note: more exactly, as of C++20 they are compiler generated from `operator<=>` ([cppref](https://en.cppreference.com/w/cpp/language/default_comparisons), [SO](https://stackoverflow.com/a/60999225))

PR #179, opened 2024-05-29
Warning fixes

pierre-guillou:
This PR contains fixes for several compiler/Doxygen warnings.

In particular:
- it adds `-Wundef` to the list of compiler warnings and fixes 4 instances when this warning was raised (using `#if` instead of `#ifdef`)
- Doxygen warnings were fixed to make sure that the documentation stays relevant to the code
- missing includes/forward declarations were added to header files to make them less dependent on the inclusion order
- the `check_code` workflow was extended to also check compilation warnings/errors in header files
- a missing `override` keyword has been added
- arguments in a `calloc` call were permuted,
- `std::shared_ptr::unique` has been replaced with `use_count` (following #148's advice)

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/check_code.yml
+++ b/.github/workflows/check_code.yml
@@ -101,8 +101,8 @@ jobs:
     - name: Use clang-check for compiler warnings
       run: |
         git ls-files \
-        | grep src \
-        | grep -E "\.cpp$|\.cxx$" \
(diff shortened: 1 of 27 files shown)


fabien-ors (approved):
Thx!

PR #186, opened 2024-06-01
Clean turning bands simulations

drenard77810:
After cleaning the Turning Band method.
The simulation method is not hosted in each ACovFunc.


NDesassis (approved):
Ok

PR #187, opened 2024-06-03
Small CMake improvements

pierre-guillou:
This PR adds small improvements related to CMake:

- the Eigen, OpenMP & HDF5 dependencies are used through CMake targets, encompassing compiler flags, include directories and linking libraries in a single `target_link_libraries` call,
- a MSVC warning about insecure C standard library calls is silenced with the `_CRT_SECURE_NO_WARNINGS` definition (we don't want to use specific Microsoft alternatives to the C standard library)
- make the Python and the R wrapper build by default when `BUILD_PYTHON` or `BUILD_R` is selected. The `python_build` and `r_build` targets are still there to build them explicitly,
- wrap `add_subdirectory(python)` and `add_subdirectory(r)` under guards. This seems more idiomatic to me than `return()` in the subdir CMakeLists.

Enjoy,
Pierre

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,8 +80,13 @@ include(cmake/install.cmake)
 ####################################################
 ## PACKAGES
 
-add_subdirectory(python)
-add_subdirectory(r)
+if(BUILD_PYTHON)
+  add_subdirectory(python)
+endif()
+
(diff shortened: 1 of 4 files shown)


fabien-ors (approved):
Thx !

PR #190, opened 2024-06-05
Make gstlearn builds reproducible

pierre-guillou:
gslearn builds are currently not reproducible: building the library in two different build directories yield libraries with different md5/sha1 sums.

This is due to the `GSTLEARN_DATE` macro, filled at configure time by CMake with a timestamp (of the configuration time). This timestamp changes each time CMake is run in any build directory.

In this PR, I propose to replace this system with information taken from the Git version control system:
- the identifier of the current commit (Git commit hash)
- and the timestamp of said commit.

As a consequence, two builds from the same Git commit should yield identical libraries.

These information are still gathered at configure time, meaning that, if a gstlearn dev creates new commits that don't require CMake to be re-run, the macros won't be updated. But I think this is still an improvement over the current situation.

Enjoy,
Pierre

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,10 +18,20 @@ option(USE_HDF5 "Using HDF5 support" OFF)
 # By default, do not activate NetCDF support
 #option(USE_NETCDF "Using NetCDF support" OFF)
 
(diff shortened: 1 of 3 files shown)


fabien-ors (approved):
Nice!

PR #191, opened 2024-06-05
Get rid of pybind11

pierre-guillou:
The function `PySys_FormatStdout` from the Python C API (https://docs.python.org/3/c-api/sys.html#c.PySys_FormatStdout) seems to follow the behavior of `pybind11::print`.

Enjoy,
Pierre


fabien-ors (approved):
Thx

PR #192, opened 2024-06-10
Use shared library objects for wrappers

pierre-guillou:
Currently, the Python an R wrappers link to a static version of gstlearn, while C++ test binaries link to the shared gstlearn library, which is the default CMake target. As a consequence, building the wrappers and the test binaries causes the compilation of both the static and the shared version of gstlearn, which is a waste of resources.

Two solutions:
1. either building the test binaries with the static library, making it the default CMake target
2. or linking the wrappers with the shared library.

Solution 1. causes link errors on Windows when building tests. Also, building with debugging symbols (`-g`) generates a 500 MB static library that will be included into each test binary. As a consequence, the build directory can grow up to 10 GB.

Solution 2. is cumbersome: there will be two shared libraries for each of the Python and R wrapper, that need to be coordinated with RPATH.

The solution proposed in this PR, which is neither 1. or 2., relies on CMake to reuse the binary objects (`.o` files) of the main gstlearn shared library for the Python and R wrappers. This way, test binaries link against this shared library, while the wrappers use the underlying objects.

This should alleviate the need of building the static library for wrappers (and accelerate a bit the CI).

Enjoy,
Pierre


fabien-ors (approved):
Nice !
Thx

PR #196, opened 2024-06-17
GitHub Actions: Force binaries for devtools

pierre-guillou:
This PR forces R to install `devtools` binary packages instead of compiling them from sources (when sources are more recent than binaries).

This should make the `publish_r_windows` workflow faster.

Enjoy,
Pierre


fabien-ors (commented):
Could you restore r_doc target as a dependency of r_build please ?

fabien-ors (approved):
Thx

PR #197, opened 2024-06-18
Enable r_doc target, switch to roxygen

pierre-guillou:
This PR enables the `r_doc` CMake target that will generate the R documentation for the wrapper methods.

The `devtools` package has been replaced with `roxygen2`, a lighter alternative (used internally by `devtools`) that seems to work better on GitHub Actions.

Enjoy,
Pierre


fabien-ors (approved):
Nice ! Thx.

PR #199, opened 2024-06-20
Switch to pyproject.toml

pierre-guillou:
This PR revamps the Python packaging by switching from `setup.py` to using a `pyproject.toml` (the `setup.py` method is deprecated and seems to be not supported anymore by recent versions of `pip`).

Static metadata inside `setup.py` have been moved inside the new `pyproject.toml` file. The `setup.py` file is still needed to force generated wheels to be platform wheels. However, instead of relying on the `wheel` package and overriding `bdist_wheel`, I propose to declare an empty `setuptools` native extension and override the building method. This only relies on `setuptools`.

To build wheels in packaging workflows, the `build` (https://github.com/pypa/build) tool is used instead of calling directly `setup.py`.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/publish_python_macos.yml
+++ b/.github/workflows/publish_python_macos.yml
@@ -67,7 +67,7 @@ jobs:
     - name: Install Python dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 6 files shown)


fabien-ors (approved):
Perfect ! Thx

PR #201, opened 2024-06-23
Covariance on sphere

drenard77810:
This branch carries the new implementation of the Spectral Simulations on the Sphere.
It also includes an improvement in the functions evalCovMatrixSymmetric(), evalCovMatrixSymmetrixOptim() and evalCovMatrixSparse(used when db1 and Db2 coincide) in order to handle the presence of Variance of Measurement error variables.


fabien-ors (approved):
More than 32 covariances implemented!

PR #209, opened 2024-07-02
Tests: Copy libgstlearn.dll into gstlearn.dll

pierre-guillou:
Following the swigex & swigex0 PRs, this PR makes a copy of `libgstlearn.dll` as `gstlearn.dll` inside the `tests` binaries directories on MSYS, following a conflict between the `r_build` and `check_cpp` CMake targets due to the use of shared objects for building the Python and R wrappers.

Enjoy,
Pierre


fabien-ors (approved):
Thx Pierre !

PR #210, opened 2024-07-02
Improve GitHub Actions workflows

pierre-guillou:
This PR cleans the GitHub Actions workflows, in particular by 
* splitting the configure, build and test steps (less log per step to check in case of an error),
* using CMake-specific environment variables
* and breaking lines around CMake configuration options (to be more diff-friendly).

Enjoy,
Pierre


fabien-ors (approved):
Many thanks Pierre!

PR #219, opened 2024-07-09
Correc bug before new branch

drenard77810:
Should correct most of the errors found by FO during the last publich.
Takes also into account some light improvements requested by MP.


fabien-ors (approved):
Thx Didier !

PR #222, opened 2024-07-10
CMake: Use BUILD_TESTING to control the compilation of test binaries

pierre-guillou:
Today, the `enable_testing()` CMake command adds a CMake option, `BUILD_TESTING`, set to `ON` by default. The C++ test binaries (directories `tests/cpp`, `tests/data` & `tests/inter`) are built through a custom CMake target, `build_tests`.

This PR leverages the `BUILD_TESTING` CMake option to control whether to build the test binaries or not. The test binaries are no longer `EXCLUDE_FROM_ALL`. This option is now set to `OFF` by default and enabled in every GitHub Actions workflow that tests them (the `nonreg` ones). An intended side effect is, in these workflows, the build step of the binaries will occur in the `Build the package` step and not in the `execute non-regression tests` step (which should be more logical).

Also, some non-ASCII UTF-8 characters were removed from the workflow files (non-breaking spaces). All workflow files are now regular ASCII files.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/check_code.yml
+++ b/.github/workflows/check_code.yml
@@ -95,6 +95,7 @@ jobs:
         cd build
         cmake \
           -DCMAKE_BUILD_TYPE=${{ matrix.opt }} \
+          -DBUILD_TESTING=ON \
           -D${{ matrix.hdf5 }} \
           $GITHUB_WORKSPACE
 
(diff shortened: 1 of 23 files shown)


fabien-ors (approved):
Thx Pierre !

Issue #149, opened 2024-04-24
Backup issue: SWIG: Python garbage collector deletes objects still used by the C++ code


fabien-ors:
## Problem
Python garbage collector deletes python objects as soon as they get out of their (python) scope, even though these objects are still referenced and used by the C++ code.

I suppose this is somehow related to how SWIG is configured?

## Example 1
```python
import gstlearn as gl
import numpy as np

dbin = gl.Db.createFromOnePoint(np.zeros(3))
dbin.setLocatorByColIdx(3, gl.ELoc.Z)
dbout = gl.Db.createFromOnePoint(np.ones(2))
model = gl.Model(gl.CovContext())
model.addCov(gl.CovAniso.createIsotropic(model.getContext(), gl.ECov.LINEAR, 1))
ksys = gl.KrigingSystem(dbin, dbout, model, gl.NeighUnique())
ksys.updKrigOptEstim(dbout.addColumnsByConstant(1, 0.0), -1, -1)
ksys.setKrigOptCalcul(gl.EKrigOpt.POINT)
ksys.isReady()  # Undefined behavior, gl.NeighUnique() does not exist anymore
```
Replacing 
```python
ksys = gl.KrigingSystem(dbin, dbout, model, gl.NeighUnique())
```
by 
```python
neigh = gl.NeighUnique()
ksys = gl.KrigingSystem(dbin, dbout, model, neigh)
```
fixes the problem, as `neigh` still exists when reaching the `ksys.isReady()` line which uses it internally

### Problem
A second example a bit more tricky: Python stores a "pointer" to a `Model` internal `ACovAnisoList` (`self._cov_list = model.getCovAnisoList()`), so `model` gets out of scope, but not `self._cov_list`.

Result: `model` is garbage collected and its internal ACovAnisoList is deleted too, leaving `self._cov_list` pointing to an invalid address.
```python
import gstlearn as gl
import numpy as np

class DualKrigingFunctor:
  def __init__(self, samples, model):
    self._samples = [gl.SpacePoint(p) for p in samples[:, :2]]
    model = gl.Model(model)  # Copy is required in my application
    neigh = gl.NeighUnique()  # Fixes Example 1 seg fault
    self._cov_list= model.getCovAnisoList()
    # Build the Dbs
    dbin = gl.Db.createFromSamples(
      nech=len(samples),
      order=gl.ELoadBy.SAMPLE,
      tab=samples.ravel(),
      names=["x1", "x2", "z1"],
[...]

PR #225, opened 2024-07-11
Turn ANeigh raw pointers into std::shared_ptr

pierre-guillou:
This PR replaces the `ANeigh` (and derivatives) naked pointers with `std::shared_ptr` smart pointers.

The `shared_ptr`s are refcounted and the memory they hold is automatically destroyed with the last `shared_ptr` instance.

This fixes Example 1 of #149.

Enjoy,
Pierre


Simon-Lopez:
My two cents (if it helps).
Turning all naked/unique pointers into shared_pointers on the C++ side can be intrusive (not to mention a possible tiny performance loss). It's not ideal if the choice of smart_pointer is linked to wrapping purposes (cf. #149).
In this cas it can be better if the reference count is managed in/by the wrapping utility (it's the purpose of pybind11/nanobind keep_alive). I do not know SWIG and I don'y know if there is the possibility to inject a low-level Py_IncRef / Py_DecRef on purposes (what keep_alive does).
There's an interesting reading here : https://nanobind.readthedocs.io/en/latest/ownership.html
And especially the proposed solution :  https://nanobind.readthedocs.io/en/latest/ownership.html#intrusive-reference-counting that mimics what Python does internally.

fabien-ors:
New approach to be designed - more transparent at C++ side.

PR #233, opened 2024-07-19
GitHub Actions: Use sccache on Ubuntu, macOS nonreg workflows

pierre-guillou:
This PR adds the support of `sccache` in the Ubuntu & macOS nonreg workflows. `sccache` will cache the object files (`.o`) generated by the compiler in a given directory. This directory will be cached across workflow runs using GitHub Actions cache (https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) automatically.

Making it work on Windows workflows is a little more involving, as `sccache` does not work well with Visual Studio solutions. I'll try to see what I can do.

Enjoy,
Pierre


fabien-ors (approved):
Thx

PR #240, opened 2024-07-23
Spatial indices

drenard77810:
End of Spatial Indices modification. A corresponding Tuto has been established in RMD.
Note that the calculation of Micro-index has been left uncoded. It will be coded when better understanding is obtained: the corresponding chunk is commented out in the Tuto file.


fabien-ors (commented):
Could you add a word about spatial indices in the CHANGES file please ?
VH::revert is never called. To be removed ?

drenard77810:
A message will be added shortly in CHANGES file.
The auxiliary methods (such as "revert") should not be suppressed in the general purpose VectorHelper class. It is there to mimic functions that are needed when translating code from R to C or from python to C. I HAD to use it to translate the R-code of Spatial Indices. Not used anymore but may be used in the future.

fabien-ors (approved):

PR #247, opened 2024-07-25
GitHub Actions: Use sccache for nonreg-tests_windows-latest-msvc

pierre-guillou:
This PR adds the support of `sccache` to the `nonreg-tests_windows-latest-msvc` workflow.

To do so, we had to switch from using the "Visual Studio" CMake generator to the "Ninja" one (which still calls `cl.exe`). The `eigen` and `boost` dependencies are fetched with `conda` (binary) rather than `vcpkg` (source).

Using the Ninja generator also forces us to give up the (now useless) static gstlearn library, since it generates a "multiple rules" Ninja error. Also, we need to skip the `Debug` & `Release` subfolders, as it interferes with Ninja/MSVC.

Enjoy,
Pierre


fabien-ors:
Thank you Pierre.
I keep this PR open during few days because I would like to be certain that 'static' target is not needed by anyone before removing it. Otherwise, is there any other solution to combine Ninja and both library flavors under windows MSVC ?
If we really remove 'static' target, we should also update the documentation ? (Makefile header, READMEs, remaining comment in cpp.cmake, etc...)

pierre-guillou:
In this version, I kept the (unused) `static` library but changed its file name as it can conflict with the `shared` library.

fabien-ors (approved):
Thank you Pierre.
Non-reg should be quicker now!

PR #249, opened 2024-07-26
Db line

drenard77810:
New class DbLine for the specific organization of samples of the Db following a set of lines.


pierre-guillou (approved):
Lgtm.
I don't know if I have enough rights to merge it though.

Issue #255, opened 2024-08-02
IDE (e.g. Visual Studio) does not list header files


remi-moyen:
Currently in Visual Studio (and presumably in other IDEs as well, given how I solved the problem), the header files are not listed in the project tree. They are correctly found when compiling or when going to where a symbol is defined, but they are not seen as being part of the project (they are seen as an external dependency instead).

This means they are not listed directly, but also that e.g. searching for "all uses" of a symbol does not search them, so this is definitely not very convenient.

The root cause is that the headers are not passed to the `add_library()` call in CMake (see e.g. [this](https://stackoverflow.com/questions/13703647/how-to-properly-add-include-directories-with-cmake) for details).

The solution is to do the same as for the source files: add a new file in `include/` that lists all the headers (I followed the same rule as in `all_sources.cmake` and listed the files rather than using `GLOB` -- I generated this file automatically by adding all `*.h*` files so hopefully it's not missing anything?), and in `cpp.cmake` add the content of this file to the `${SOURCES}` variable.

fabien-ors:
I confirm that headers are not listed in Visual Studio (or Code::Blocks) projects generated by CMake. On the other hand, there is no problem while using VS Code or Eclipse CDT. The reason is explained in this answer:
https://stackoverflow.com/questions/13703647/how-to-properly-add-include-directories-with-cmake/74409439#74409439
I am not an expert of Visual Studio, but is there any plugin in Visual Studio that offer the possibility to handle CMake project (gstlearn is a CMake project)?

remi-moyen:
Your link is part of the same page as the one I posted so I guess we found the same solution!

Visual Studio handles CMake projects correctly without any plugin (or rather, CMake can generate Visual Projects correctly when running e.g. `cmake -G "Visual Studio 17 2022"` though this is the default on Windows so this isn't really needed). But, as the link mentions, the headers need to be in the `${SOURCES}` for them to be listed as part of the project itself (rather than as external dependencies).

So my fix was to do exactly that, adding the headers to `${SOURCES}`. This works for me (Visual Studio) and I don't think this should have any negative effect in other development environments? But I haven't tested (on Linux I just use the command line, no IDE).

[...]

PR #256, opened 2024-08-02
IDE (e.g. Visual Studio) does not list header files #255

remi-moyen:


remi-moyen:
This PR was not the right way and the proper (hopefully?) fix is now in PR #271.

I had to make another branch and PR because I messed up the `issue255` branch while trying to update it.

PR #259, opened 2024-08-13
Schur

drenard77810:
The new class KrigingCalcul encompasses all the Algebraic operations needed in Kriging with all possible options:
- Standard Krigings: simple, ordinary, Universal
- Multivariate
- With or without drift (possibly external)
- With Bayesian hypotheses.
Still to be faced:
- correction for Bayesian in multivariate
- additional options such as Collocation and Cross-validation (specifically in Unique Neighborhood)


pierre-guillou (commented):
What's the purpose of the newly added `include/Estimation/2`, `include/Matrix/2`, etc. files?

drenard77810:
Hi Pierre
These are spurious files created by VSCode. They should not be included in the distribution.
A better view would be to understand when and where they are created and to correct this artifact.

pierre-guillou (approved):
lgtm

PR #260, opened 2024-08-13
Matern

drenard77810:
The Covariance named BESSEL_K is now named MATERN.
The covariance BESSEL_J is renamed as BESSELJ


NDesassis (approved):
Ok

PR #263, opened 2024-08-20
Optmization simbool

drenard77810:
This operation is meant to decrease the computing time for SIMTUB.
In the reference bench mark, we succeeded decreasing the number of operations (valgrind) from 22.931 to 14.577, hence a reasonably impressive improvement.

The main gain is obtained by an optimization in VectorT. Note that this improvement (suppressing some tests) does not impair the robustness of gstlearn. It will also impact (speeding up) all the different procedures. Hence a good and general improvement.


fabien-ors (changes requested):
Could you remove these spurious files please ?

bench_VGrid.out
src/Calculators/Result.out
src/Core/Result.out
src/Db/Result.out
tests/data/Result.out
tests/rmd/output/Test_Vario.out

drenard77810:
Cleaning done

PR #264, opened 2024-08-22
Spde multivar phase2

NDesassis:
The aim of this branch is to use Eigen::VectorXd instead of VectorDouble in the algebraic computations in SPDE.

Note1 : some tests results change (up to some decimals) and a lot of ref files had to be updated.
Note2: some copies are still unnecessary (even if the copy from VectorDouble to Eigen;:VectorXd are compulsory at the interfaces. To avoid unnecessary copy, one should use Eigen::Map<Eigen::VectorXd> when Eigen::VectorXd are expected as output argument. It requires to use a template type of Eigen (not necessary for Eigen::Map<const Eigen::VectorXd> for the inputs.
Note 3: some files have to be cleaned (indentation, suppresion of commented code, ...)


fabien-ors (approved):
Merge OK.
Note: A fix is still needed for ScaleOp::_addToDest.

PR #272, opened 2024-08-29
ProjMatrixMulti

NDesassis:
New classes have been created to handle projections in the spde multivariate framework (matrix free version and matrix sparse version):
**ProjMulti
ProjMultiMatrix**

They are fully functional as demonstrated by the test 
**tests/py/test_ProjMultiMatrix.py** 

A method 

**static std::vector<std::vector<const ProjMatrix\*>> create(std::vector<const ProjMatrix\*> &vectproj, int nvariable);**

has been written to create the argument to easily instantiate the class **ProjMultiMatrix** in the classical spde multivariate case.

 But the classes can handle more general ways to perform multivariate projections (for instance for FKA filtering) or to perform seismic inversion.

Some other classes have been created (to handle the non matrix free case) but work is still in progress.
**SPDEOpMatrix
PrecisionOpMultiMatrix**


fabien-ors (changes requested):
Thank you ! Here are 2 small "change requests" :
1. There is a strange "return" in MatrixSparse::addProdMatVecInPlaceToDest line 600. Could you confirm ?
2. Could you remove the file "ScaleOp copy.cpp" ?

fabien-ors (approved):

PR #277, opened 2024-09-02
Fix compilation warnings

remi-moyen:
This is a batch of small changes to fix compilation warnings. I've tried to group similar changes in each commit to make it easy to check them. Since it's so small I haven't created an issue, hopefully that's not... an issue!

The warnings were triggered on Windows but they're actually not specific to Windows, it's mostly about unused function arguments.


remi-moyen:
Commit 627ab38 is causing one test to fail because the function now prints some more messages than before.

This was one commit I wasn't entirely sure about. I think my code is correct and the expected output of the test needs to be updated, but maybe you want to remove the `message()` calls instead?

fabien-ors:
The failures for non reg seems to occur in target r_doc.
I have added some verbosity in this target to try to diagnose the issue.

fabien-ors:
I have also removed the messages in DbGraphO

include/Covariances/ACovFunc.hpp
```diff
@@ -54,7 +54,7 @@ class GSTLEARN_EXPORT ACovFunc : public AStringable
   virtual bool         hasSpectrumOnSphere() const { return false; }
   virtual bool         hasSpectrumOnRn()     const { return false; }
   virtual bool         hasMarkovCoeffs()     const { return false; }
-  virtual double       normalizeOnSphere(int n = 50, double scale = 1.) const { return 1.;}
+  virtual double       normalizeOnSphere(int /*n*/ = 50, double /*scale*/ = 1.) const { return 1.;}
```
pierre-guillou, line 57: roxygen2 seems to dislike these default arguments with commented variable names.

fabien-ors:
@remi-moyen : Could you please roll back your commits #798ce11 and #f9ba5da (remove the comments around the unused arguments) and prefer using the following macro (in geoslib_define.h) : DECLARE_UNUSED.
(an example is available in ACovFunc::simulateTurningBand)
I suggest you to have a look to the Style.hpp documentation where some C++ tricks are detailed for developers.
Thank you very much for your kind contributions.

remi-moyen:
I've pushed a new commit that uses `DECLARE_UNUSED()`.

remi-moyen:
`-Wunused-parameter` is explicitly turned off in `cpp.cmake`, I think this is a bad idea?

This would catch most of my changes above, and while I agree that it's annoying to fix those warnings when we develop code, it helps keeping the code cleaner (and once in a while they catch a real bug!). Forcing developers to write `DECLARE_UNUSED` also makes things clearer to both the person writing the code, and those reading it later. In some cases this may lead you to question whether a parameter should even be there, which again is a good thing!

So I suggest to remove the `-Wno-unused-parameter` flag.

fabien-ors:
Thank you Rémi for all !!
I have removed -Wno-unused-parameter in cpp.cmake and resolved some conflict.
I will be able to merge this PR this evening, I guess !!! (non-reg is running)

PR #279, opened 2024-09-03
Precision op multi no stat

NDesassis:
New classes to handle SPDE in multivariate case.
Mainly PrecisionOpMulti (for the matrix-free operator) and PrecisionOpMultiMatrix for the (sparse) matricial case.
Both inherit from ALinearOp and implement _addToDest but also simulateInPlace.


fabien-ors (approved):
Nice Nico Thx !

PR #286, opened 2024-09-17
Reduce memory copies with Eigen

pierre-guillou:
This PR tries to reduce the number of copies when using Eigen. In particular, copies happen when converting between `Eigen::VectorXd`, `VectorDouble` and `VectorEigen`.

This PR leverages `Eigen::Map<Eigen::VectorXd>` to avoid unnecessary copies, in particular in `LinearOpCGSolver`, `AMatrixDense` and `AMatrixSparse`.

A lot more copies might be lurking in the code, especially when methods taking `Eigen::VectorXd` are passed `Eigen::Map` arguments. This is at least a beginning.

Pierre


NDesassis (approved):
Thx!

PR #287, opened 2024-09-18
Factor some allocations in MeshETurbo

pierre-guillou:
While doing debug, I ran [heaptrack](https://github.com/KDE/heaptrack) on `test_SPDEAPI` and it showed me memory allocations that were (partly) easy to factor.

This PR focuses on the `MeshETurbo` methods. There was a `VectorInt` copy in `_getPolarized` that I replaced with a const ref. Other methods were using a `VectorInt indg` that I refactored using a global variable (I lacked a better idea).

These two improvements yield the following results in `test_SPDEAPI`:

| Metric            | Before | After |
|-------------------|--------:|-------:|
| Exec time (s)     | 23     | 11    |
| #Allocations      | 75M    | 28M   |
| #Temp allocations | 18M    | 4M    |
| Memory cons (MB) | 82 | 82 |

There is still room for improvement, namely in the handling of `VectorNum`s: if one vector is modified with a `use_count() > 1`, a vector copy is triggered before the modification. No idea yet on how to tackle that though.

I've not tested the impact on the other test executables (apart from checking their output on GitHub).

Enjoy,
Pierre

### DIFF
--- a/include/Mesh/MeshETurbo.hpp
+++ b/include/Mesh/MeshETurbo.hpp
@@ -120,7 +120,7 @@ public:
 private:
   int  _defineGrid(const VectorDouble& cellsize);
   void _setNumberElementPerCell();
-  int  _getPolarized(VectorInt indg) const;
+  int  _getPolarized(const VectorInt &indg) const;
   int  _addWeights(int icas,
                    const VectorInt &indg0,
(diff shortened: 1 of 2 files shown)


NDesassis (approved):
Thx!

PR #288, opened 2024-09-18
Cpp20

NDesassis:


fabien-ors:
I cancel this PR (we are going to add C++ 20 compatibility in a branch forked from dev directly)

PR #293, opened 2024-09-20
fix matern bug (K-Bessel)

NDesassis:
- Changes have been made for seed handling in the SPDE simulation functions.
- Replace the call to **besselk** of gstlearn by the function **std::cyl_bessel_k** to avoid numerical instabilities for large values of h. This function should replace besselk directly.
Modification of the latex formula to have a more standard expression. Suppression of the scale parameter in the expression (it should be done in all the formulas).


fabien-ors (changes requested):
Could you add a sentence in CHANGES to explain the bug and indicate that it has been fixed ?

NDesassis:
> Could you add a sentence in CHANGES to explain the bug and indicate that it has been fixed ?

Done

fabien-ors (approved):
Thank you Nicolas !

PR #305, opened 2024-09-24
Spde : use of std::span<double> instead of Eigen::VectorXd.

NDesassis:
The aim of this PR was to make the code less dependent of the Eigen library and to avoid copies when one work on slices of vectors.
For this purpose, we used the std::span<T> class thanks to the recent migration to C++ 20.
We suppressed the VectorEigen class and we now only use Eigen::VectorXd in the Matrix classes and in the ALinearOp class to be able to use the solvers of Eigen (e.g Conjugate Gradient solver).


src/LinearOp/IProjMatrix.cpp
```diff
@@ -11,73 +11,48 @@
...
-    Eigen::VectorXd myOut(outv.size());
-    
-    // Assume outv has the good size
-    int error= point2mesh(myInv, myOut);
-    
-    Eigen::Map<Eigen::VectorXd>(outv.data(), outv.size()) = myOut;
-    return error;
+  outv.resize(getApexNumber());
+  constvect myInv(inv.data(), inv.size());
+  vect myOut(outv);
```
pierre-guillou, line 30: Here, you are forced to define local variables of type `std::span` to pass them by reference to `point2mesh`. You can avoid that by passing `std::span` by copy and calling `point2mesh(inv.getVector(), outv.getVector()); `, using the implicit conversion from `std::vector` to `std::span` or using temporaries: `point2mesh(constvect{inv}, vect{outv})`

src/Matrix/AMatrixDense.cpp
```diff
@@ -92,6 +92,12 @@ double AMatrixDense::_getValueByRank(int irank) const
   return *(_eigenMatrix.data() + irank);
 }
 
+std::span<const double> AMatrixDense::getColumnPtr(int icol) const
+{
+  const auto a = _eigenMatrix.col(icol);
+  int n = getNRows();
+  return std::span<const double>(a.data(),n);
```
pierre-guillou, line 99: I hope `a` is a non-owning pointer inside `_eigenMatrix`, otherwise this method might trigger a use-after-free.

pierre-guillou (approved):
This PR looks very good, we can get rid of a lot of `Eigen::VectorXd`, `Eigen::Map` and all of `VectorEigen`.
While this can I think be merged as it is, I have two small concerns:
- there is some small code duplication with methods that accept either `VectorVectorDouble` or `std::vector<std::vector<double>``. Maybe a template parameter might help to factor that?
- several newly added methods take `std::span` as a reference. When references are `const`, users can pass temporaries as arguments, whereas with non-`const` reference parameters, users are forced to create variables (c.f. IProjMatrix.cpp). Passing `std::span` by value instead of by reference helps removing explicit conversions.

What do you think?
Pierre

doc/demo/python/Test_SPDEPGS.ipynb
```diff
@@ -542,7 +542,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.5"
+   "version": "3.10.12"
```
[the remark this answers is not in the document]
NDesassis: I had to run this test to understand a bug (linked to this PR). I cleared the output but it changed the python version. I don't know why...

include/Basic/VectorHelper.hpp
```diff
@@ -52,6 +52,8 @@ class GSTLEARN_EXPORT VectorHelper
   static double maximum(const VectorDouble &vec, bool flagAbs = false, const VectorDouble& aux = VectorDouble(), int mode=0);
   static double minimum(const VectorDouble &vec, bool flagAbs = false, const VectorDouble& aux = VectorDouble(), int mode=0);
   static double maximum(const VectorVectorDouble &vec, bool flagAbs = false);
+  static double maximum(const std::vector<std::vector<double>> &vec, bool flagAbs = false);
+
```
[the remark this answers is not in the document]
NDesassis: I totally agree with this comment. I am not really familiar with template. Could we envisage to report this task for a future PR?

src/LinearOp/IProjMatrix.cpp
NDesassis: Ok to supress references. I remarked that an automatic conversion was done but are we sure that there is no copy (remember the case of Eigen::Map<Eigen::VectorXd> given in place of Eigen::VectorXd, it worked but you told me that a copy was made).

PR #309, opened 2024-09-27
GitHub Actions: Remove 10 min idle timeout on sccache daemon

pierre-guillou:
The `nonreg-demos-courses_ubuntu_latest` GitHub Actions workflow has been displaying zeroed `sccache` statistics in the workflow run summary. After several tries, I understood that `sccache`, when idle for 10 minutes (as written in its documentation), stops its daemon. The affected workflow does a lot of testing around Python and R, these take more than 10 minutes, leading to the zeroed `sccache` stats.

This PR sets the `SCCACHE_IDLE_TIMEOUT` environment variable to 0 (no timeout) to fix this issue.

Enjoy,
Pierre


fabien-ors (approved):
Nice Pierre !

PR #310, opened 2024-09-27
Pass std::span by value

pierre-guillou:
Since `std::span` are non-owning pointers, they can be passed by value instead of by reference.

This PR converts all uses of `std::span` as parameters to pass-by-value. It can help to shorten function calls, in particular with `VectorDouble`s (c.f. src/LinearOp/IProjMatrix.cpp): C++ seems to forbid passing temporaries as const references, but passing them by value is OK.

Enjoy,
Pierre


NDesassis (approved):
Thanks a lot!

PR #317, opened 2024-10-03
GitHub Actions: Fix conda bug in nonreg Windows MSVC

pierre-guillou:
I don't know why the previous one-liner doesn't work anymore but this fix should.

Pierre


fabien-ors (approved):
Thx Pierre

PR #318, opened 2024-10-03
Avoid heap allocations of litterals with std::string_view

pierre-guillou:
`std::string_view` is to `std::string` what `std::span` is to `std::vector`. `std::string_view` is a C++17 feature that has a similar API to `std::string` but is constant and doesn't own the underlying buffer. In particular, `std::string_view::substr` doesn't allocate.

In gstlearn, a few methods that take a `String &` parameter are passed character literals as arguments: the Enums `fromKey` methods are a good example. Doing so will allocate a copy of the literal on the heap whereas `std::string_view` will point directly to the data segment.

In this PR, I use `std::string_view` instead of `String` around Enums. Explicit conversions to heap-allocated `Strings` are performed if needed be.

Since Enum methods are exposed to the R & Python wrappers, Swig typemaps have been introduced to convert `std::string_view` to or from heap-allocated `String`s under the hood.

Enjoy,
Pierre

### DIFF
--- a/include/Basic/String.hpp
+++ b/include/Basic/String.hpp
@@ -20,8 +20,8 @@
 
 GSTLEARN_EXPORT void skipBOM(std::ifstream &ins);
 
-GSTLEARN_EXPORT String toUpper(const String &string);
-GSTLEARN_EXPORT String toLower(const String &string);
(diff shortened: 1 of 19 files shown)


NDesassis (approved):
Thx!

PR #320, opened 2024-10-07
Modification in Covariance

NDesassis:
The aim of this PR is to open the possibility to create more general covariance functions as the Gneiting or separable functions. The Gneiting has been added to show an example.
Some modifications on the covariance classes (ACov and ACovAnisoList) have been made to make the separation between (matricial) correlation functions and sills matrix (in order to open to more general multivariate covariance functions).


fabien-ors (approved):
Some improvements should be done later to prevent memory copies while projecting a  SpacePoint into a sub-space

PR #321, opened 2024-10-08
Cholesky refactoring

drenard77810:
Cholesky is now a new branch with ACholesky and two daughters CholeskySparse and CholeskyDense.
The Matrix classes MatrixSparse and MatrixSquareSymmetric do not hold any dependency with Cholesky class not methods anymore.

Note that the two internal commands "compare_test" and "update_test" which allow handling the output of the non-regression files have been translated into BASH


include/Estimation/KrigingSystem.hpp
```diff
@@ -255,6 +256,7 @@ class GSTLEARN_EXPORT KrigingSystem
   MatrixSquareSymmetric _priorCov;  // Dimension NF * NF
   VectorDouble          _postMean;
   MatrixSquareSymmetric _postCov;
+  CholeskyDense         _postCovChol;
```
NDesassis, line 259: Are we sure that it will always be dense? evalCovMatrix has a sparse version (useful with tapering).

include/LinearOp/MatrixSquareSymmetricSim.hpp
```diff
@@ -30,45 +30,25 @@ class Cholesky;
...
   virtual ~MatrixSquareSymmetricSim();
 
-  const AMatrix* getMatrix() const {return _mat;}
-  int getSize() const override { return _mat->getNRows();}
-  bool isSparse() const {return _sparse;}
-  bool isInverse() const {return _inverse;}
+  const AMatrix* getMatrix() const;
+  int  getSize() const override;
+  bool isEmpty() const { return _factor == nullptr; }
+  bool isSparse() const { return _sparse; }
```
NDesassis, line 41: Do you really need to know if it is sparse?

NDesassis (approved):

PR #323, opened 2024-10-10
GitHub Actions: Update publish workflows to be compatible with C++20

pierre-guillou:
This PR updates the Ubuntu & Windows `publish` workflows to switch to more recent versions of said OSes.
Namely, Ubuntu 20.04 is updated to 22.04 and Windows 2019 to 2022.

Note that Ubuntu 20.04 and Windows 2019 are still in use in the `coverage` workflows. Depending on the availability of the C++20 features inside the C++ standard library, we may need to upgrade them too.

Note 2: the macOS 12 jobs of the `publish` workflows seem to take a quite large amount of time (more than 2h) t. Warning messages show that macOS 12 is not supported anymore by brew. As a consequence, LLVM is built from source instead of downloading binaries.

Pierre

### DIFF
--- a/.github/workflows/publish_python_ubuntu.yml
+++ b/.github/workflows/publish_python_ubuntu.yml
@@ -41,8 +41,8 @@ jobs:
             {py: "3.11"},
             {py: "3.12"}
           ]
-        # Only one old Linux and a generic platform name BUILD_PLAT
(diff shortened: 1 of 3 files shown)


fabien-ors (commented):
Thx Pierre !!

fabien-ors (approved):

PR #324, opened 2024-10-10
Reduce allocations around MeshETurbo, Grid & Rotation

pierre-guillou:
Following #287, this PR reduces the number of allocations shown by [heaptrack](https://github.com/KDE/heaptrack) when running `test_SPDEAPI`.

Several techiques were used:
- replacing or complementing methods taking `VectorInt &` or `VectorDouble &` by corresponding methods taking `std::span<int>` or `vect`,
- replacing `VectorDouble` local variables or class members with `std::vector<double>`,
- using global `std::vector` variables to reuse storage,
- using Swig's `%extend` to make sure the correct `Grid` methods are used in the C++ code but not in the wrappers.

This table shows the evolution of `heaptrack`'s metrics on `test_SPDEAPI`:

| Metric            | Before | After |
|-------------------|-------:|------:|
| Exec time (s)     | 12     | 8     |
| #Allocations      | 26M    | 12M   |
| #Temp allocations | 3.1M   | 3.0M  |
| Memory cons (MB)  | 82     | 82    |

There's still a lot of allocations remaining, especially around matrices. Some will be hard to remove (those pertaining to Eigen), others less so (`VectorDouble`). I'll try to work on the latter.

Pierre

### DIFF
--- a/include/Basic/Grid.hpp
+++ b/include/Basic/Grid.hpp
@@ -77,10 +77,12 @@ public:
   VectorDouble getCellCoordinatesByCorner(int node,
                                           const VectorInt& shift = VectorInt(),
                                           const VectorDouble& dxsPerCell = VectorDouble()) const;
+#ifndef SWIG
   double indiceToCoordinate(int idim0,
-                            const VectorInt& indice,
-                            const VectorDouble& percent = VectorDouble(),
-                            bool flag_rotate=true) const;
+                            const constvectint indice,
+                            const constvect percent = {},
+                            bool flag_rotate        = true) const;
+#endif // SWIG
   VectorDouble indicesToCoordinate(const VectorInt& indice,
                                    const VectorDouble& percent = VectorDouble()) const;
   void indicesToCoordinateInPlace(const VectorInt& indice,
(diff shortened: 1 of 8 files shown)


fabien-ors (commented):
Do you suggest that methods which exist in two versions (one with VectorDouble arguments and another with vect arguments) should not be duplicated directly in the C++ class?
In other words, do we have to keep the method using std::span in the C++ class and use %extend in swig_inc.i (to wrap the method using VectorDouble for Python and R users)?
Currently, in this PR there are the two logics: Grid::indiceToCoordinate which use %extend, and Rotation::rotateDirect which is duplicated in Rotation class.

pierre-guillou:
Duplicating the methods in the `Grid` class doesn't work very well due to automatic conversions from `std::vector` to either `VectorDouble` or `std::span`. The compiler doesn't know which implementation to choose.
For `Rotation`, the new methods are only called in the `Grid::getCoordinates` methods since I changed the type of the `_work2` member to `std::vector`.

But since the `Grid` methods are also used through the wrappers, we need to provide conversion methods from `VectorDouble` to `constvect`. Thus the `%extend`.

This raises deeper questions about the performance of the C++ API: should we continue to use `VectorDouble` and pay the price of the `std::shared_ptr`, or try to use data structures with less overhead (`std::vector`, `std::span`) and push the use of `VectorDouble` to the SWIG code? Is there a simple way in SWIG to generate `VectorDouble` methods that call the corresponding `std::vector` or `std::span` methods or should we intensively `%extend`?

fabien-ors (approved):

PR #325, opened 2024-10-11
Fix ref for VarNN macos nonreg

fabien-ors:
Fix the Var3/6/9/11/12/13/14 reference logs for macos
Delete specific msvc and msys logs for Var3. They were identical to linux


fabien-ors:
The reference log for PCA/MAF is for ubuntu only. The github version is different from our local linux.
We just need to put our local reference in myoutput.

PR #328, opened 2024-10-14
CMake: Fix include_directories BUILD_INTERFACE

### LINKED ISSUE #326: Generated CMake package contains hard-coded paths to the source code
When installing gstlearn with `make install` (or the INSTALL target in Visual Studio), a CMake package is generated with file `<install_dir>/share/gstlearn/cmake/gstlearnTargets.cmake`. This allows using `find_package()` and then just the target `gstlearn::shared` in other libraries using it.

However `gstlearnTargets.cmake` contains a reference to the path to the source code, in my case `D:/Dev/sandbox/gstlearn`:
```CMake
set_target_properties(gstlearn::shared PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "D:/Dev/sandbox/gstlearn/;D:/Dev/sandbox/gstlearn/;${_IMPORT_PREFIX}/include/gstlearn;${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Eigen3::Eigen;iphlpapi;rpcrt4"
)
```

This is because in `cpp.cmake` the `target_include_directories()` are specified as `PUBLIC` which means they are part of the "interface" of the target (see [CMake's doc](https://cmake.org/cmake/help/latest/command/target_include_directories.html)).

The issue is that when the package is then copied to a computer that does not have the source code (i.e. those paths are invalid), CMake fails to load the package:
```CMake
CMake Error in /data/<snip>/src/Core/CMakeLists.txt:
  Imported target "gstlearn::shared" includes non-existent path

    "D:/Dev/sandbox/gstlearn/"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not
  provide.
```

Setting the `target_include_directories()` to `PRIVATE` fixes this as the generated `gstlearnTargets.cmake` becomes:
```CMake
set_target_properties(gstlearn::shared PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/gstlearn;${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Eigen3::Eigen;iphlpapi;rpcrt4"
)
```

So this is the change I have made in my version.

However in `cpp.cmake` there is a comment saying that there is a good reason for the includes to be `PUBLIC` (for tests), so I don't know if setting them `PRIVATE` will cause issues there. Still, the way they are currently set is definitely wrong as we end up with hard-coded paths in the package.

pierre-guillou:
Following #326, this PR applies the proposed changes by @remi-moyen.
Quotes have been kept for safety.

Fix #326.

Pierre

### DIFF
--- a/cmake/cpp.cmake
+++ b/cmake/cpp.cmake
@@ -119,9 +119,9 @@ foreach(FLAVOR ${FLAVORS})
   # PUBLIC is mandatory for tests and packages (no need to install)
   target_include_directories(${FLAVOR} PUBLIC
     # Add includes path for compiling the library
-    $<BUILD_INTERFACE: ${INCLUDES}>
+    "$<BUILD_INTERFACE:${INCLUDES}>"
     # Add binary directory to find generated version.h and export.hpp
-    $<BUILD_INTERFACE: ${PROJECT_BINARY_DIR}>
+    "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>"
   )
 
   # Set some target properties
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -24,7 +24,7 @@ endif()
 # Include directories
 target_include_directories(shared PUBLIC
   # Installed includes are made PUBLIC for client who links the shared library
-  $<INSTALL_INTERFACE:include/${PROJECT_NAME}>
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>"
 )
 
 # Install the shared library


fabien-ors (approved):
Thank you Pierre !
Could you also apply the fix to swigex[0] too please?

PR #332, opened 2024-10-18
Nlopt

NDesassis:
This branch add a test of optimization using nlopt.


cmake/cpp.cmake
```diff
@@ -172,6 +186,9 @@ foreach(FLAVOR ${FLAVORS})
     target_link_libraries(${FLAVOR} PUBLIC -liphlpapi -lrpcrt4)
   endif()
 
+  #Nlopt include 
+  target_include_directories(${FLAVOR} PRIVATE ${NLopt_INCLUDE_DIRS})
+
```
pierre-guillou, line 191: Due to the `NLopt::nlopt` in `target_link_libraries` that should carry the dependency include directories, this `target_include_directories` should be useless and safely removed I think.

NDesassis: Exact. I've done the modification. Thank for your comment.

PR #336, opened 2024-10-23
Execute nonreg demos/courses in parallel in GH actions

fabien-ors:
This PR should improve the execution time of the demos/courses non regression :-)
In the dedicated github action, The demos/courses for R and Python are executed in parallel (the two target languages in parallel)
Enjoy ^^


fabien-ors:
Last commit should not has been done here.
Wait for matrix_typemap PR before mergining

PR #340, opened 2024-10-31
GitHub Actions: Add test jobs to publish workflows

pierre-guillou:
This PR adds `test` jobs to the `publish-r` and `publish-python` workflows. These test jobs do some basic testing on the packages generated by the `build` jobs before they are uploaded to `pypi` or `CRAN`. 

In fact, these are copies of the `coverage` workflows that have been put here to prevent faulty jobs to be uploaded.

Doing so in separate jobs makes testing independent on the build environment and thus helps detecting missing dependencies in gstlearn packages.

Corresponding workflow run: https://github.com/pierre-guillou/gstlearn/actions/runs/11608521333

Enjoy,
Pierre


fabien-ors (approved):
Thx Pierre

PR #345, opened 2024-11-05
Schur for dual

drenard77810:
This PR is meant to terminate the definition of the KrigingCalcul class.
As a recall, this class is able to handle any Kriging configuration, i.e.:
- any number of variables (i.e. Kriging or Cokriging)
- Simple, Ordinary or Universal Kriging
and handle special cases such as:
- Bayesian hypothesis on drift coefficients (note that the Bayesian hypothesis on the Covariance parameters is not handled)
- Cross-validation: given a fixed set of data (Unique Neighborhood), there is a special trick for deriving the Cross-validation (using the k-fold option [leaving more than one point out]) while saving lots of calculations
- Collocated Cokriging: given a fixed set of data (Unique Neighborhood), and knowing the values of one of several variables at the Target point, there is a special trick for performing the estimation of the remaining variables at this target. Note that the collocated option has never been implemented in previous in-house software!
- Dual writing: given a fixed set of data (Unique Neighborhood), a trick enables calculating the estimation (for one or several variables) in a quicker way using the Dual writing

All these options are tested in a CPP test names test_Schur.

### DIFF
--- a/3rd-party/csparse/csparse_f.h
+++ b/3rd-party/csparse/csparse_f.h
@@ -36,6 +36,7 @@ Website: https://gstlearn.org
 #ifndef _CS_F_H
 #define _CS_F_H
 
+#include <cstdio>
 #include <stdlib.h>
 #include <stddef.h>
 #include "csparse_d.h"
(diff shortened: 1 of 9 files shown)


fabien-ors (approved):
Nice!

PR #346, opened 2024-11-05
Add dependency to NLopt

pierre-guillou:
This PR takes over #332 to provide support to the [NLopt](https://nlopt.readthedocs.io/en/latest/) library inside gstlearn.
Since the library is not header only, either a shared library is needed at runtime or gstlearn is built with a static NLopt library. For now, NLopt is only used in a test source file so it is not exposed to users of gstlearn's C++ API.

The `nonreg` and `publish` workflows have been updated to take NLopt into account. For the `nonreg` workflows, the system package system is used. For the `publish` workflows, a static build of NLopt is performed, either through `vcpkg` on Windows or with a custom composite action for Ubuntu and macOS.

Corresponding `publish` workflow run: https://github.com/pierre-guillou/gstlearn/actions/runs/11685081744

Pierre

### DIFF
--- a/.github/actions/nlopt_static_unix/action.yml
+++ b/.github/actions/nlopt_static_unix/action.yml
@@ -1,0 +1,24 @@
+name: 'NLopt static Unix'
+description: 'Build & install NLopt static libraries on Ubuntu & macOS'
+runs:
+  using: "composite"
+  steps:
+
+    - name: Build & install nlopt static libraries
+      shell: bash
+      run: |
+        wget https://github.com/stevengj/nlopt/archive/v2.7.1.tar.gz
+        tar xvfz v2.7.1.tar.gz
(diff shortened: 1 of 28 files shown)


fabien-ors (changes requested):
Thank you very much Pierre

1/ Please fix a typo in the root README: In the Nlopt section: "Adapt the *Eigen* source code path... etc"
2/ Please add a comment for NLopt_ROOT variable in the header of the Makefile
3/ Please add the NLopt credit to the root README in the section "Credits"
4/ In the same file, add the * `-DNLopt_ROOT="path/to/nlopt"` in "Important Notes" section

fabien-ors (approved):
Thx Pierre

fabien-ors:
And Thank you Nicolas ;)

PR #347, opened 2024-11-06
Amelioration KrigingCalcul

drenard77810:
Amelioration KrigingCalcul (supprimer pointeur et correction fuite memoire). Fix #345


fabien-ors (approved):
Nice cleanup!
The concept could be generalized :-)

PR #348, opened 2024-11-07
For suprema

drenard77810:
This branch contains the slight adjustments required by SUPREMA project (while coded within gstlearn ... essentially in segy subproject)


fabien-ors (approved):
Only whitespace

PR #351, opened 2024-11-08
Branch to code extract diag for a sparse matrix

drenard77810:
This branch has been developed in order to implement the feature of extracting the diagonal of a sparse matrix corresponding to a PrecisionOp.
We must remember that a PrecisionOp is based on a matrix Q=S^n where S is the sparse matrix contained in a LinearOp.
The Sparse Matrix S is sparse and can be stored (even in the case of a 3D data set). Whereas the matrix Q (of PrecisionOp), although it is sparse, may contain too many non-zero terms to be explicitely stored.
 
This is the reason why PrecisionOp is the matrix-free version and PrecisionOpCs is its implementation where Q is explicitly stored.
The method extractDiag() applied on Q has been implemented in order to retrieve this diagonal:
- directly when Q is stored explicitly (simple use of sparse matrix method)
- in the matrix-free case where Q is not actually stored.
-


fabien-ors (approved):
Thx Didier

PR #353, opened 2024-11-12
New class for Model_Optiom

drenard77810:
A new class has been designed for the fitting of a Model.
For the time being, it is designed only for fitting from a Variogram.
The non-regression test test_nlopt.cpp has been enhanced to demonstrate the capbility of this new fitting procedure (it will replace gradually the one based on FOXLEG).
The next step will be to create a fitting procedure based on Maximum Likelihood (in the same branch).


fabien-ors (changes requested):
Could you had some words about main API changes in CHANGES file please? krigingSPDE, simulateSPDE, logLikelihoodSPDE, etc...

fabien-ors (approved):
Thx Didier !

Issue #203, opened 2024-06-24
Db.dumpToNf and Db.createFromNf are very long


NDesassis:
Shouldn't be stored in a more clever format than ascii? Binary? Hdf5?

PR #355, opened 2024-11-13
Improve performance around serialization

pierre-guillou:
This PR provides several performance improvements around the serialization of `DbGrid`.

@NDesassis provided me with the following Python script:

```py
import gstlearn as gl
db = gl.DbGrid.create(nx = [200,200,200])
db.dumpToNF("toto.csv")
```

Using sampling profilers (such as [samply](https://github.com/mstange/samply)) I have been able to reduce the time spent in this script by a factor of 2.

Most of the performance gains were due to avoiding buffer flushes (`std::endl`). The remaining were obtained by factoring allocations (replacing `VectorNum`s with `std::vector`s / `std::span`s to avoid the cost of `std::shared_ptr`, pass by reference instead of returning locally-allocated vectors, using global variables to reuse storage, etc.).

Now a large majority of the serialization time is spent into `operator<<(double)`.

Here are some Samply/Heaptrack metrics:

|Metric|Before|After|
|----------|----------|-------|
| Exec time (samply) | 15s | 7.5s |
| Exec time (heaptrack) | 48s | 12s |
| #Samples  (samply) | 15k | 7.3k |
| Mem consumption | 272 MB | 272 MB |
| Allocations | 88081640 | 81651 |
| Temp. allocations | 8042940 | 42948 |

This is a first step for #203.

Pierre

### DIFF
--- a/include/Basic/ASerializable.hpp
+++ b/include/Basic/ASerializable.hpp
@@ -75,7 +75,7 @@ protected:
   template <typename T>
   static bool _recordWriteVec(std::ostream& os,
                                const String& title,
-                               const VectorT<T>& vec);
+                               const std::vector<T>& vec);
 
   template <typename T>
   static bool _recordRead(std::istream& is,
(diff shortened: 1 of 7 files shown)


fabien-ors (approved):
Thx Pierre. To be generalized to other serializations.

PR #356, opened 2024-11-15
GitHub Actions: Fix link issue with NLopt on Windows

pierre-guillou:
This PR fixes link issues with NLopt on GitHub Actions Windows VMs (c.f. for instance #353).

The solution consists in building NLopt ourselves instead of relying on package systems. Static libraries should be preferred to shared ones, at least for the Python wrapper. A new composite action has been introduced to take care of that. It is quite similar to the one used for Ubuntu and macOS, we can merge the two if needed be.

For the `nonreg-tests_windows-latest-rtools` workflow, switching from vcpkg to using pacman-provided NLopt packages seems to overcome the issue in `create_doc.R`.

Pierre


fabien-ors (approved):
Thx Pierre

PR #359, opened 2024-11-20
Goulard refactor

drenard77810:
This code creates a new class ModelOptimSills dedicated to fit the sill matrices of a Model which may contain one or several covariances for a one or several variable case, for any space dimension. This corresponds to the well-known Goulard option.

It is a first step before including this Sill Fitting (on option) in a general Model fitting procedure (the sill being fitted using Goulard) based on nlopt. Note that, optionally, nlopt can be used even to fit the sills.


fabien-ors (changes requested):
Could you add some words about NLopt new dependency in the CHANGES file please ?

fabien-ors (approved):
Nice Didier ! Thx you

PR #361, opened 2024-11-21
Fix Windows compilation error due to a missing include.

### LINKED ISSUE #360: Windows compilation error (VS2017)
I have had this error for a long time and fixed it in my development version but for some reason I never committed it.

Compilation fails in FileLAS.cpp, line 30:
`{ return std::tolower(c);});`

This is because it is missing the include of `<cctype>`. This works on Linux, I guess that include gets added indirectly somehow, but adding it shouldn't cause any issue.

remi-moyen:
This fixes #360


fabien-ors (approved):
Thx Rémi !

PR #363, opened 2024-11-21
Add a preprocessor variable GSTLEARN_VERSION_NUMBER in version.h.

### LINKED ISSUE #362: Add preprocessor variable for gstlearn version
There is a variable GSTLEARN_VERSION defined in version.h however this is a string ("1.5.0") which means it cannot be used in preprocessor logic. For example, one cannot write `#if GSTLEARN_VERSION > "1.4.1")` or similar.

I propose to add numeric variables to make this possible (for example if one wants to have code that compiles with several versions of gstlearn). There are two ways this can be done:
- The most complicated way (but also the most robust) is to do the same as e.g. Qt (at the beginning of [qglobal.h](https://github.com/mburakov/qt5/blob/master/qtbase/src/corelib/global/qglobal.h)) and define several variables VERSION_MAJOR, VERSION_MINOR etc. as well as macros to do the proper comparisons. This is a bit of work and probably overkill.
- The more straightforward idea is to just compute a single variable (which I've called GSTLEARN_VERSION_NUMBER) which is major*10000+minor*100+patch so e.g. 1.5.0 becomes 10500. This will break if one of the numbers goes above 99 and makes it a bit less convenient to check for ranges of versions (e.g. "all 1.5 but neither 1.4 nor 1.6" needs two comparisons).

We use the 2nd approach in other projects and in practice this works perfectly, so this is what I'm suggesting here. Note that the first one can always be added as well, the two are not incompatible. But the 2nd one is enough for me.

remi-moyen:
The variable GSTLEARN_VERSION cannot really be used for anything else than displaying it because it's a string (so we cannot do e.g. #if GSTLEARN_VERSION == "1.5.0"). We could pass the individual bits of the version as defined by CMake (major, minor, patch), but my experience is that it is convenient to wrap them in a single number.

GSTLEARN_VERSION_NUMBER is built as MAJOR*10000 + 100*MINOR + PATCH which leaves space for up to 99 minor/patches, so in practice it usually fine.

This fixes #362

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,8 @@ execute_process(
 string(TOUPPER ${PROJECT_NAME} PROJECT_NAME_UP)
 
 # Fix version.h automatically
+# Define a numerical version number (usable in e.g. #if ( VERSION >= ...))
+MATH( EXPR gstlearn_VERSION_NUMBER "${gstlearn_VERSION_MAJOR} * 10000 + ${gstlearn_VERSION_MINOR} * 100 + ${gstlearn_VERSION_PATCH}" )
 configure_file(version.h.in version.h)
 
 # Detect presence of multi configuration generators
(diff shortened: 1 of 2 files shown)


fabien-ors (approved):
Thank you Remi for this contribution and for the related issue !

Issue #366, opened 2024-11-21
Keep compatibility with C++17


remi-moyen:
gstlearn now sets the C++ standard to C++20 in cpp.cmake. This is an issue for me as on Windows I only have Visual Studio 2017 that does not support C++20.

I would accept that this is a problem for me and that I need to update my toolchain, however it turns out that, at least for now, only a couple of simple changes are required to keep being compatible with C++17, so the attached PR is doing exactly that.

The two things are:
- The main reason for which C++20 was used is to have access to `std::span` but we can use `boost::span` instead which just requires a not-too-old version of Boost and works with C++17. So the main change here is in `geoslib_define.h` to use `boost::span` instead of `std::span`. To ensure this is restricted to compilers that don't support C++20, this is only triggered if a new CMake variable `USE_BOOST_SPAN` is passed to CMake. This variable is not defined by default so if you don't do anything, `std::span` will still be used as before.
- A minor thing is that `std::map::contains()` was added in C++20, and is used in one file. Here I've just replaced those calls by `std::map::count()` which is a bit less efficient and a bit less clean but given the code I don't think it matters, and again this allows keeping the compatibility with C++17 so I really would like to see this.

I understand that on the long term other features of C++20 will end up being used (and at one point I will probably update my toolchain!). However if this is feasible with relatively little changes, as is the case so far, I will try and maintain the compatibility with C++17 for as long as I can.

remi-moyen:
As an aside, `cpp.cmake` contains `set(CMAKE_CXX_STANDARD_REQUIRED True)` which should cause an error with a compiler that does not support C++20. Obviously this is not the case with Visual Studio 2017 as I can compile with it.

This may be because VS 2017 supports parts of C++20 (with the flag `/std:c++latest`) even though it does not have full C++20 support. Or maybe it's because Visual Studio does not properly support `CMAKE_CXX_STANDARD_REQUIRED` (this was apparently the case for some very old versions of VS, maybe this is still the case?).

In any case, combined with the changes I made to stick to C++17, I think this line is a bit contradictory. I'm not suggesting to remove it or change it in any way, but it's just something I noted and that is a bit weird.

PR #367, opened 2024-11-21
Keep compatibility with C++17

remi-moyen:
This fixes issue #366 (at least with the current state of the code, I know this might need more work in the future...).


fabien-ors (changes requested):
Thank you Rémi !
Could you add some words about this new configuration variable in the root README file please ?

remi-moyen:
Good point. I've added an item in the "Important notes" section.

remi-moyen:
I just noted an issue with my fix, the `#define USE_BOOST_SPAN` is not passed to projects that use gstlearn.

The fix is to use `target_compile_definitions(... PUBLIC ...)` instead of `add_definitions()` (or `add_compile_definitions()`), but this requires setting this individually on each project (whereas `add_definition()` is global!).

I'll make another commit with that change.

Issue #368, opened 2024-11-22
Make dependency to NLopt private


remi-moyen:
At the moment the dependency to NLopt in `cpp.cmake` is public:
`target_link_libraries(${FLAVOR} PUBLIC NLopt::nlopt)`

This means that projects that want to use gstlearn through the CMake configuration files will indirectly need to know where NLopt is. However, NLopt is explicitely compiled as a static library which means that actually code using gstlearn doesn't need to know anything about it.

So it would be better to make that dependency PRIVATE.

PR #369, opened 2024-11-22
Link to NLopt as PRIVATE so that projects using gstlearn don't need NLopt

remi-moyen:
Since gstlearn uses the static library version of NLopt, there is no need for projects using gstlearn to know anything about it, so it can be a PRIVATE dependency.

This fixes issue #368

### DIFF
--- a/cmake/cpp.cmake
+++ b/cmake/cpp.cmake
@@ -170,7 +170,7 @@ foreach(FLAVOR ${FLAVORS})
   target_link_libraries(${FLAVOR} PRIVATE Boost::boost)
   
   # Link to NLopt
-  target_link_libraries(${FLAVOR} PUBLIC NLopt::nlopt)
+  target_link_libraries(${FLAVOR} PRIVATE NLopt::nlopt)
 
   # Link to HDF5
   if (USE_HDF5)
--- a/tests/cpp/CMakeLists.txt
+++ b/tests/cpp/CMakeLists.txt
@@ -69,6 +69,12 @@ foreach(TEST_SOURCE_FILE ${TEST_SOURCES})
     add_dependencies(${TEST_NAME} prepare_cpp)
     # Trigger the build of the test with the target build_tests
     add_dependencies(build_tests ${TEST_NAME})
+
+    # test_nlopt.cpp explicitly calls NLopt so it must link with it (gstlearn::shared
+    # links it as PRIVATE to hide it from projects using it).
+    if("${TEST_NAME}" STREQUAL "test_nlopt")
+      target_link_libraries(${TEST_NAME} PRIVATE NLopt::nlopt)
+    endif()
 endforeach(TEST_SOURCE_FILE ${TEST_SOURCES})
 
 # Display test output in case of failure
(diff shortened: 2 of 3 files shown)


fabien-ors (changes requested):
Hi Rémi,
This proposition doesn't seem to work for programs that want to link with nlopt.
In the non-regression checks, the test program test_nlopt.cpp cannot link to gstlearn due to nlopt symbols.
Please could you verify what is the issue?
Does we need to modify our tests CMakeFiles?
Thank you for your contribution.

remi-moyen:
You are right that this doesn't work with programs that link with NLopt, or rather that those programs need to explicitly add a dependency to NLopt.

In a sense, this is the goal of my proposed change, because at the moment all code that uses gstlearn must know where NLopt is (i.e. `find_package(NLopt)` must work). This adds a constraint on my code as I cannot just copy the installation of gstlearn, I must also copy the installation of NLopt. And this constraint is not really needed since gstlearn uses the static library of NLopt (and none of the headers of gstlearn include NLopt headers), so my code doesn't really need to know that NLopt even exists.

Now you are also entirely right that my change is breaking the test test_nlopt.cpp, I didn't check that. Sorry.

The solution that I'm suggesting (I'll try making another commit for that, if I don't mess up again with updates...) is indeed to change the CMakeLists.txt for the tests. test_nlopt.cpp explicitly calls NLopt and therefore I don't think it's a big problem to have it explicitly link with NLopt. This adds a special case for it in the CMakeLists.txt but that doesn't seem like a huge issue to me?

remi-moyen:
Hopefully my commit will be correct once the branch is updated to dev?

I'm still confused as to why I get a few other unrelated commits pushed when I added my new commit (these are not all commits added since I created the branch, just those from... another branch that I created myself?), but I think once the branch is up to date those should no longer be seen as specific to this branch.

Sorry if I'm doing it the wrong way, please tell me if I should follow another workflow for updating a PR branch when I need to add another commit later?

pierre-guillou:
Hi Rémi,
If you want, I can help you by removing the unrelated commits in your branch.
You can do that locally with `git rebase origin/dev` (with `origin/dev` up-to-date or after `git fetch origin`) then `git push --force-with-leave origin issue368` to propagate to GitHub.
I can do that for you if you don't feel comfortable (and also fix the non-working tests).
Pierre

pierre-guillou:
In fact, the classic way to update your branch with more recent commits from `origin/dev` is to `git merge origin/dev` into your branch. You can also do that locally then push (without `--force-with-lease`) to GitHub. No need to PR your own branch.
The rebase technique is a little bit more complex at first (it rewrites history by moving the commits from your branch on top of `origin/dev`, thus the need of `--force-with-lease`) but it simplifies the global history after the merge (less tangled history).

For the CI failure, I think you need to add `OR "${TEST_NAME}" STREQUAL "test_ModelFitting"` in the tests/cpp/CMakeLists.txt since `test_ModelFitting.cpp` also includes `nlopt.h`.

pierre-guillou:
Or remove the `#include <nlopt.h>` inside `test_ModelFitting.cpp` since it does not seem to be used in the body of the test…

remi-moyen:
Yes, I forgot that other test. I think it's best to remove the `#include` from there, I don't see any reason to have it.

fabien-ors (approved):

PR #380, opened 2024-12-07
Shift op optim

drenard77810:
New class ShiftOpStencil


fabien-ors (approved):
Nice Didier & Nico! Thx

PR #383, opened 2024-12-11
Shift op optim

NDesassis:
This PR aims to plug the ShiftOpStencil approach for SPDE.
In this first version, the PrecisionOp and PrecisionOpMulti has to be created with the option stencil = true (default = false) in order to use the stencil approach.


fabien-ors (approved):
Thx Didier and Nicolas

PR #384, opened 2024-12-15
Modifs presentation brgm

drenard77810:
These modifications have been induced by the preparation of the Talk for BRGM


fabien-ors (approved):
Thx Didier

PR #385, opened 2024-12-17
CI: Fix building rgl for publish_doc workflow

pierre-guillou:
Add missing `libpgn-dev` package to make `rgl` build correctly.

Corresponding Action run: https://github.com/pierre-guillou/gstlearn/actions/runs/12369216682

Pierre


fabien-ors (approved):
Good point Pierre !

PR #386, opened 2024-12-17
Use matrix in api when possible

drenard77810:
The API for adding a covariance offers the possibility to define the sill as a Matrix (rather than with VectorDouble).
The previous methods have been kept with an extension OldStyle.


NDesassis:
Nice PR with a clear purpose.

NDesassis (approved):
Nice PR with a clear purpose.

Issue #316, opened 2024-10-02
Wrong SPDE results when some input points are outside the grid


remi-moyen:
I'm doing a 2D kriging with SPDE and get very wrong results when some of my input data points are outside of the grid.
![image](https://github.com/user-attachments/assets/46036b4c-bba0-492e-8e44-6d990f75a719)

My data set has one data point that is (very narrowly) outside of the grid. When running SPDE, it correctly reports that "Sample #1 does not belong to the grid" but the results are wrong. A "normal" kriging (with `gl.kriging()`) is fine.

If I remove that data point, the results are correct. But there is more to it, because various other ways to edit the input data seem to fix it. For example, if I move the data point to the end of the file, then the result is correct. Also, a few of my data points have a value that's very different from the rest (-999.99, these are actually no data values but it doesn't matter, the kriging should be able to use them even if it's a user error!). If I remove those points, then the result looks correct. I also noted that the problem in the results look like if a very low value (say, -999.99!) was used instead of the correct value in a few places.

Putting all those observations together, I suspect that the issue is that a sample is skipped, but then some internal array doesn't update its indexes correctly so samples after that one are "shifted" and mixed. That probably is happening somewhere in `MeshETurbo::resetProjMatrix()` and in particular I am suspecting this bit of code:

https://github.com/gstlearn/gstlearn/blob/97dc47ec90fb96b935ab90e4f4907898159cde6a/src/Mesh/MeshETurbo.cpp#L544-L557

The `nvalid` variable counts the number of valid samples, but when a sample is outside of the grid the `continue` is triggered after `nvalid` has been increased. So I think the increment of `nvalid` should be moved below that test? But maybe there are other places where skipped samples should be correctly accounted for?

I can share the data set as well as a minimal Python example to reproduce if needed.

remi-moyen:
I think this issue can now be closed as it is indeed solved.

I did not specifically test with the same data again, however I had another data set with points outside the grid where I was getting wrong results with gstlearn 1.4.1. Removing the points outside the grid fixed the issue, which confirms that this was the same bug as this one. Then I switched to the dev version of gstlearn and got the correct results, which confirms the fix.

Thanks!

fabien-ors:
Nice !
Thank you for your feedback.

PR #387, opened 2024-12-18
Correction bug SPDE API

drenard77810:
This fixes the issue #316.
It checks the consistency of results beween SPDE and Kriging, for subsequent cases:
- when the data is masked oof
- - when the data contains NA
- - when the data is located outside the Meshing internal grid


fabien-ors (approved):
Thx Didier

PR #388, opened 2024-12-18
Cosmetics in publish workflows

fabien-ors:
Just some cosmetics in order to make github action workflows as close as possible to those from swigex.


NDesassis (approved):
Ok

PR #396, opened 2025-01-10
Make some Python dependencies optional

pierre-guillou:
This PR makes the Python packages `geopandas`, `plotly` and `shapely` optional dependencies of the `gstlearn` Python package.

Now, a `pip install gstlearn` won't install those dependencies. Users that want to use the `gstlearn.plot` module should use `pip install gstlearn[plot]` to force the installation of the dependencies, which have been grouped inside the `plot` optional feature.

Other non-essential Python dependencies can be added to this dependency group or into another one. Tell me if I should include/exclude other packages.

Pierre

### DIFF
--- a/doc/courses/python/01_gstlearn_start.ipynb
+++ b/doc/courses/python/01_gstlearn_start.ipynb
@@ -57,7 +57,7 @@
    "outputs": [],
    "source": [
     "# Uncomment next line for installing last release of gstlearn package (remove '#' character)\n",
-    "#!pip install gstlearn"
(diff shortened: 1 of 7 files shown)


fabien-ors (changes requested):
Thank you Pierre.
Please, could you :
- add "[plot]" to the "pip install gstlearn" command in the 01_gstlearn_start.ipynb script ?
- Fix typo in README : alonside => alongside.

Thank you !

fabien-ors (approved):

PR #400, opened 2025-01-21
Fusion2

drenard77810:
Some additional tests have been performed in order to prepare the migration of KrigingSystem to its new-style version (not yet finished)


NDesassis:
Accepted as a step but cleaning is needed. (example : setOrigin(...) const with mutable origin in ASpace is not a good idea.

NDesassis (approved):

PR #414, opened 2025-02-12
Various renames

drenard77810:
This branch (which will last a very short time) is meant to perform all the renaming phases that were necessary.
In particular:
- modify 'npas', 'dpas', 'ipas' into 'nlag', 'dlag' and 'ilag'
- normalize the names of the non-regression files (cpp, py, r, ipynb and rmd).

One exception is left unmodified: test_nlopt.cpp (which should have been renamed to test_Nlopt.cpp).
But doing so introduces a bug that is not easy to understand. Therefore the old name is kept


fabien-ors (approved):
Nice :)

PR #418, opened 2025-02-13
CI: Switch to pixi to handle both Python & native dependencies

pierre-guillou:
This PR switches the `nonreg-tests_windows_latest_msvc` workflow to use `pixi` instead of `conda` + `pip` to download Python and native `gstlearn` dependencies.

This workflow is the only one currently using `conda`. The latest `conda` release (25.1) seem to break down with the current CI setup, in particular around the `setup-conda` action. Switching to `pixi` fixes this issue.

Pierre


fabien-ors (approved):
Thx you very much Pierre.
Could you also push this commit on the other branches that are currently under PR ?

PR #425, opened 2025-02-21
Improvement for remi moyen

drenard77810:
In this last commit a special file cpp/template_Kriging.cpp allows constructing the Kriging:
- either using the API
- or using covariance and drift parts, using the KrigingAlgebra level.


fabien-ors (approved):
Thx Didier

PR #429, opened 2025-02-27
Optim de acov

drenard77810:
This branch (which should be closed as soon as the tests are performed successfully) is meant to establish a milestone in the cleaning of the following sets of classes:
- ACov
- KrigingSystem
- KrigingAlgebra

Future developments:
- ACov:
We should certainly consider some extra cleaning phases in the ACov class in order to clean some spurious remaining functions (@ND). This framework should then be put in action for specific cases such as the "variable and gradient" cokriging or the Kriging in the Potential model (which is in reality a cokriging of Z, its gradient components and possible some tangent components). These two examples are the first ones where the CovProportional class cannot be used anymore).
-KrigingSystem and KrigingAlgebra:
Close the cleaning of KrigingSystem suppressing chunks of code that are not reachable anymore (for example, code for Image Neighborhood should have been moved to CalcImage; code for global should have been moved to CalcGlobal).
Add some functions of interest similar to those already added such as exhibiting the "Mu" and/or the "Lambda"...

Peripheral improvement: using the new KrigOpt class to cimpilfy all the API of kriging. Warning: we must update all the ipynb and the rmd consequently.


fabien-ors (approved):
Need urgently to find a clean way to handle Vector* pointers as function argument through swig (see TODO in swig_inc.i) tp prevent the memory leak.
Maybe use std::shared_ptr<VectorDouble> in gstlearn ?
@pierre-guillou

PR #431, opened 2025-03-05
Krigeage gneiting

drenard77810:
This PR must be performed, followed by a merge within DEV.
This is just a step before we are able to perform Kriging using a new type of Generic Model (for Gneiting type of Covariance for example).
The next step is to review the whole set of classes considered as calculators. At a given stage, those methods are based on Model which should become a ModelGeneric. Then all the relevant methods should be updated (if possible).

When this second phase is acheived, then we will pretend that we can perform Kriging with these new types of Models.


drenard77810:
Some more information.
This branch is clearly not finished. I use the same branch for the second phase...

NDesassis (approved):

PR #434, opened 2025-03-06
Remove deprecated USE_HDF5 code

pierre-guillou:
This PR removes the old HDF5 code around `HDF5format.cpp`/`HDF5format.hpp` that exposed the HDF5 API to gstlearn consumers. The corresponding CMake/Makefile macros have also been cleaned.

TODOs have been put into `GibbsMulti.cpp`. In a future PR, serialization will be implemented for this class.

Enjoy,
Pierre


fabien-ors (changes requested):
Please, could you also cleanup comments in non-reg / publish workflows (HDF5_ROOT, etc...)

fabien-ors (approved):
Nice!
Thx Pierre

PR #441, opened 2025-03-07
Model: Fix memory leak

pierre-guillou:
`setCovAnisoList` is cloning its pointer parameter…

Enjoy,
Pierre


fabien-ors (approved):
Thx Pierre

PR #459, opened 2025-03-13
Replace static global vectors in MeshETurbo, Db with mutable class members

### LINKED ISSUE #445: Remove static local variable in MeshETurbo.cpp
In src/Mesh/MeshETurbo.cpp, there is a local variable declared as static:

https://github.com/gstlearn/gstlearn/blob/fd6be45af4eb12355617e0904f58d1591e2d2128/src/Mesh/MeshETurbo.cpp#L141

This is used in a lot of functions such as:
https://github.com/gstlearn/gstlearn/blob/fd6be45af4eb12355617e0904f58d1591e2d2128/src/Mesh/MeshETurbo.cpp#L178-L180

I suspect the reason for this static variable is to avoid allocating a new variable every time, for performance reasons. However I think this is a bad practice as this makes this code slightly harder to read but more importantly it makes it very thread-**unsafe**, and I wonder how important the performance gain really is (i.e., is it worth breaking thread-safety?).

This is thread-unsafe because if two threads happen to call functions of this class the same variable may be used by both simultaneously. Having the variable as static makes it worse as this can happen even if using two entirely unrelated instances of MeshETurbo, without any obvious reason for the user (you may expect a single MeshETurbo to not be thread-safe but there is no reason for two different MeshETurbo to interact!).

So at the very least, I recommend changing this variable to make it a (mutable) class member. This would still not be thread-safe but only if the same MeshETurbo is used in two threads. Different instances of MeshETurbo would become independent from each other. This would have no impact on the (possible?) optimisation of reusing the same variable.

In fact, doing so would ensure that the initialisation (`indg.resize(ndim);`) performed in all functions is almost always a no-op (the dimension of a mesh never changes!). This would in turn improve performances if using simultaneously several MeshETurbo instances of different dimensions: the current code will re-dimension the vector all the time, whereas a class member would never need to be re-dimensionned more than once.

But I wonder if the performance gain of using a single variable is really that noticeable? Since using a single variable breaks thread-safety, I would prefer not to have it unless it has been shown to make a real difference, and if so then this should be clearly documented? So unless there has been tests showing this is worth it, I would suggest to remove the variable entirely and re-define it in each function using it.

(I stumbled upon this because my compiler was warning me about variables with the same name shadowing this global variable in e.g., this code:
https://github.com/gstlearn/gstlearn/blob/fd6be45af4eb12355617e0904f58d1591e2d2128/src/Mesh/MeshETurbo.cpp#L207
this warning can be fixed trivially by renaming the variable (or ignored altogether) but I think it's best to fix the global variable, which I personally see as an issue)

pierre-guillou:
This PR replaces the static global `std::vector` variables in `MeshETurbo.cpp` and `Db.cpp` introduced in #287 and #324 with mutable class member variables.

No change observed in the number of memory allocations.

Fix #445.

Pierre

### DIFF
--- a/include/Db/Db.hpp
+++ b/include/Db/Db.hpp
@@ -966,4 +966,7 @@ private:
   std::vector<int> _uidcol;   //!< UID to Column
   VectorString _colNames;    //!< Names of the variables
   std::vector<PtrGeos> _p;   //!< Locator characteristics
+
+  /// factor allocations
+  mutable std::vector<int> uids;
 };
--- a/include/Mesh/MeshETurbo.hpp
+++ b/include/Mesh/MeshETurbo.hpp
@@ -158,4 +158,11 @@ private:
   bool  _isPolarized;
   Indirection _meshIndirect;
   Indirection _gridIndirect;
+
+  /// factor allocations
+  mutable std::vector<int> indg;
+  mutable std::vector<int> indices;
+  mutable std::vector<double> lambdas;
+  mutable std::vector<double> rhs;
+  mutable std::vector<int> indgg;
 };
--- a/src/Db/Db.cpp
+++ b/src/Db/Db.cpp
@@ -651,8 +651,6 @@ void Db::setArrayByUID(const VectorDouble& tab, int iuid)
   }
 }
 
-static std::vector<int> uids;
-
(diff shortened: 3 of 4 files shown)


fabien-ors (approved):
Nice!

PR #462, opened 2025-03-13
Point d'etape dans acov optim

drenard77810:
This covmat is a stone in the direction of optimization.
I focussed on evalCovOptimByRank whcih computes the point covariance matrix between two data sets.
It is the basic example of calculation.
I provided 3 implementations for testing the progress (you can access giving to the value OptimCovmat by OptCustom mechanism:
1 - old version with loop on db1 (data) THEN loop on db2 (target)
2 - the new version where loops are interchanged. We measure some time gain
3- the version where I suppressed any reference to VVI to evacuate possible time loss. The proof is that VVI does not cost extra time (and it offers the possibility to work with heterotopy and selection easily). As a consequence VVI should be maintained (possibly improved to allow changing the order of the ' nested loops).
The last version has been instrumented so as to calculated the time for each chunk of this very linear writing mode of the function. It allows evaluating which part takes more time...

The next step is to reorder covmatsym and covmatTarget by inverting the order of the loops. This will then but the starting point for umltimate optimization, trying to find the times of the past, or even the ones of ND.

I will start this next step after the current PR is merged... just ot start on fresh and good basis.


fabien-ors (approved):
Thx Didier

PR #476, opened 2025-03-20
Correcting isoline on rotated grid

drenard77810:
The representation of isolines on a rotated grid has been fixed.


fabien-ors (approved):
Thx Didier

PR #477, opened 2025-03-21
Python: Move scipy, pandas to optional dependencies

### LINKED ISSUE #442: Dependencies of the gstlearn python package
Currently, the package declares dependencies to the following libraries: numpy, scipy, pandas, matplotlib, plotly. Although they are all useful for "scripting" applications, we only require a few of them when using gstlearn as a computational library.

It would be nice to have a "core" gstlearn package that comes with only a restricted set of "computational" dependencies, and to declare the others as optional configurations in the pyproject.toml.

At first sight, for the core dependencies, I would say:
- numpy: required
- pandas: unsure, it is used a few times in `pygstlearn.i`, so it could be kept for now
- scipy: not needed. It is used only once in `pygstlearn.i`. It would be nice to find an alternative so it can be removed from '"mandatory" dependencies
- matplotlib: not needed
- plotly: not needed

This also poses the question of handling "specific" python module dependencies (segyio, ipywidgets, ...)
An example of `pyproject.toml`:
```toml
[project]
...
dependencies = ["numpy", "pandas"]  # Core computational dependencies
...
[project.optional-dependencies]  # Can be install using pip install gstlearn[all]
common = ["matplotlib", "plotly", "scipy"]  # For test, demos, ...
segy = ["segyio"]
gui = ["ipywidgets", ...]
all = ["gstlearn[common,segy,gui]"]  # Require pip >= 21.1
all = ["matplotlib", "plotly", "scipy", "segyio", "ipywidgets", ...]  # Alternative
```

pierre-guillou:
This PR creates a new optional dependency group for the Python package, named `conv`. This group controls the installation of `scipy` and `pandas`, two dependencies of the gstlearn Python package.

A few functions use these dependencies, they have been moved to a separate Python module that is automatically loaded if the dependencies are installed. If not, an error message is displayed within the `toTL` C++ function. This concerns the `Vario`, `Db` and `Table` conversions from/to Pandas Dataframes and several types of gstlearn matrices to NumPy or scipy sparse.

Also, the Python function `Db_fromPanda` has been renamed to `Db_fromPandas` to fix a typo (`pandas` has a trailing 's').

To manage all optional Python dependencies, a third dependency group, `all`, is introduced in this PR. `pip install gstlearn[all]` will install all optional dependencies (`matplotlib`, `scipy`, `geopandas`, etc.).

Closes #442.

Pierre

### DIFF
--- a/doc/courses/python/01_gstlearn_start.ipynb
+++ b/doc/courses/python/01_gstlearn_start.ipynb
@@ -194,7 +194,7 @@
     "\n",
     "* Some classes of the *gstlearn* C++ library have been extended in Python:\n",
     "  * Almost all classes are 'stringable' (those which inherit from `AStringable`, [see here](https://soft.mines-paristech.fr/gstlearn/doxygen-latest/classAStringable.html)), that means that you can type the object name in the Python console prompt and hit 'Enter' key to obtain a detailed description of the object content. The same output text is obtained using the `display` method (e.g. `mygrid.display()`)\n",
-    "  * Some classes have an additional Python method named `toTL` (i.e. 'to Target Language') that permits to convert an object into the corresponding Python type. For example, the instruction `df = mygrid.toTL()` permits to create a pandas `DataFrame` from a `Db` object. In that case, the newly created `DataFrame` will contain all variables from the Db (but locators and grid parameters (for DbGrid) will be lost)\t"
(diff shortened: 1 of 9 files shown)


fabien-ors (approved):
Nice Pierre ! Thx U.

PR #481, opened 2025-03-23
Add %newobject for ::clone methods (to prevent memory leaks)

fabien-ors:
Prevent memory leaks after a call to clone method in Python or R. Created object will be garbage collected if not used anymore. To prevent this, the object has to be stored in a variable.


NDesassis:
Nice

NDesassis (approved):
Nice!

PR #482, opened 2025-03-23
First steps for simplification of API Kriging and others using KrigOp…

drenard77810:
First steps before Kriging API is simplified.
As a matter of fact, most of these functions will, in a near future, use the generic argument KrigOpt.
For the time being, this new class has been developed.
The next step will be to modify the prototype of kriging, xvalid, simtub, ...

So please, do not close this branch, even after success


fabien-ors (approved):
Thx Didier.

PR #483, opened 2025-03-25
Remove indirections in AMatrixDense, CholeskyDense

pierre-guillou:
This PR add two improvements in handling Eigen variables as class members:
- in `AMatrixDense`, the variable `_eigenMatrix` with type `std::unique_ptr<Eigen::Map<Eigen::MatrixXd> >` has been changed to `VectorDouble`. An `Eigen::Map` is generated on the fly when needed.
- in `CholeskyDense`, the `_factor` member has been switched from a pointer to a variable. This eliminates some memory leaks in the copy constructor and the copy assignment operator at the expense of a deep copy.

Enjoy,
Pierre


NDesassis (approved):
Thanks a lot!

PR #484, opened 2025-03-26
Remove shared_ptr inside VectorT

pierre-guillou:
The `std::shared_ptr` inside `VectorT` has a confusing behavior and a negative effect on performance. This PR removes the `std::shared_ptr` indirection. There are small gains visible in the non-regression tests timings: on Ubuntu, the whole CI step goes from 22min to 13min. This is particularly true for the `bench_Tree` benchmark, whose execution time goes from 350s to 90s on Ubuntu and from 60s to 25s on Windows.

Enjoy,
Pierre


fabien-ors (approved):
Super Nice !
Thank you Pierre

PR #489, opened 2025-03-31
Python: Add request, IPython optional dependencies to package

pierre-guillou:
This PR adds `requests` and `IPython` as optional dependencies of gstlearn's Python package. They are required by `document.py`. A new optional dependency group, `doc`, is introduced to contain the two of them.

Pierre


fabien-ors (approved):
Thx Pierre

PR #490, opened 2025-04-02
Fix bug in Driftist to prevent crash with setMean

fabien-ors:
The _means vector for DriftList class was wrongly resized to 0 size when deleting all drifts. Its length should be fixed to NVariables during the whole life of the DriftList object.


pierre-guillou (approved):
Rebase + push --force seems good

PR #492, opened 2025-04-02
New graphics matplotlib

drenard77810:
This is BIG modification as I developed a new version of PLOT.PY (Python version of the graphics that are provided to help the user playing with gstlearn in a graphic mode).
All files had to be updated to cope with the new syntax .... which, I beleive, offers much more robustness and possibilities.


NDesassis (approved):
Nice!

PR #503, opened 2025-04-15
Fix compilation with C++17

remi-moyen:


NDesassis (approved):

pierre-guillou:
Thanks for your fix!

Given that `std::any_of` is also valid C++20, I suggest to make your proposal the default one, rather than keeping two competing code paths. I'll push this next.

remi-moyen:
A lot of `std::ranges` functions just replace equivalent `std` functions that took iterators. Since the ranges make the syntax simpler, I kept that code, but of course adding the `#ifdef` makes the code more complex also (but it makes it clear to someone looking at that code that there is a good reason to *not* use ranges!).

Anyway, I'm probably the only one still not using C++20 (and I'm hoping to drop it one day, sadly not yet...), so as long as it compiles for me, I don't care which approach is used.

Thanks for handling it, and for humouring my old compiler 😉 !

Issue #505, opened 2025-04-15
Ordinary kriging with SPDEOp


remi-moyen:
Currently the code in `SPDEOp` does not work if trying to compute drifts (for example for ordinary kriging). There are two functions `computeDriftCoeffs()` and `evalInvCov()` that are used for that (in the old code) but they don't work.

In addition to fixing those two, I think some more work is needed at a higher level (e.g., in `SPDE`?) to make it easy for the user to add drifts.

I've submitted PR #506 for fixing `computeDriftCoeffs()` as currently calling it crashes because of a wrong size used. But that's just avoiding a catastrophic failure, it's not fixing trends properly.

Comparing with the old code, I think in `evalInvCov()` what's missing is just to implement the equivalent of the commented-out line:
```c++
//VectorHelper::multiplyConstant(w2s,-1);
```
which can be done for example with:
```c++
for (auto& i : w2s)
  i *= -1;
```
(but maybe a better solution is to add a function in `VectorHelper` to handle that).

NDesassis:
Solved.

PR #506, opened 2025-04-15
Fix a crash in SPDEOp::computeDriftCoeffs()

remi-moyen:
This code was likely copy-pasted from the old version PrecisionOpMultiConditional where the second argument (drifts) was a VectorVectorDouble. Now it's a matrix so using the total size is wrong.

This commit is part of #505 (but not enough!).


NDesassis (approved):
Thx Remi.

NDesassis (approved):
Thx Remi.

fabien-ors:
This PR should fix #505

Issue #496, opened 2025-04-04
Add thread separation of random number generation


remi-moyen:
At the moment when random numbers are needed (for example for a simulation), they are generated through a global object:
https://github.com/gstlearn/gstlearn/blob/00c2fc545aa0b5df117d177c77387f0b33a68cd7/src/Basic/Law.cpp#L24

There is a function to control the random seed used by the generator:
https://github.com/gstlearn/gstlearn/blob/00c2fc545aa0b5df117d177c77387f0b33a68cd7/src/Basic/Law.cpp#L63-L71

However when running code in parallel, this means all threads are using the same RNG and the values drawn in each thread will thus vary according to how threads are scheduled.

I would like each thread to use its own RNG to ensure reproducibility of results.

One solution (but I'm not sure this is enough!) is to create an RNG in each call to functions that create full random vectors (e.g., in `VectorHelper::simulateGaussianInPlace()`) and add an argument for the seed in those functions. Then bubble up this seed argument up to where it's needed in the API, for example up to `SPDEOp::simCond()`.

This "leaks" the abstraction of the RNG into some higher-level code but I think that's OK. It should not surprise a user calling `simCond()` that there is an `rng_seed` argument. I don't think there should be any performance impact either, creating an `std::mt19937` object should be quick when using it to draw a large random vector.

Another solution could be to store the global RNG into some thread-local variable but that's probably actually harder to code properly and would also complicate setting the seed in different threads to ensure reproducibility.

remi-moyen:
While rooting around in that code I noticed the `law_set_old_style()` function. It doesn't seem to be called from anywhere in the main library (it's called in a couple of tests), which seems to mean the code is by default using the "old" RNG. That one seems entirely thread-**unsafe** given that the same variables (`Random_factor` etc.) are used everywhere, so different threads are going to mix things in that generator!

So at the moment I think by default the RNG is not working in multiple threads. If possible I would prefer if the default was changed to the "new" RNG (`std::mt19937`) as it is at least thread-safe, if not thread-reproducible.

drenard77810:
Hi Remi
Just a short comment on your proposal.
I am aware that the OLD STYLE random value generator are not thread safe...
This is one reason why we appreciated using the NEW STYLE version using std::mt** which is much safer.

HOWEVER, we noticed a problem in this new technology: it is not cross-platform safe.
[...]

Issue #495, opened 2025-04-04
Improve thread-safety of simulations with SPDEOp


remi-moyen:
I'm trying to speed-up generating many simulations by having them done in parallel. Essentially, the code I would like to run is (freeing of all objects is ignored for simplicity!):

```C++
Model* model = ...;
MeshETurbo* mesh = ...;
PrecisionOpMulti* precision = new PrecisionOpMulti(model, { mesh }, true /* or false i.e., stencils or not */);

ProjMulti* proj = ...;

MatrixSquareSymmetricSim* invnoise = ...;

SPDEOp* op = new SPDEOp(precision, proj, invnoise, precision, proj);

std::vector<double> conditioning_data = ...;

#pragma omp parallel for
for (int s = 0; s < nb_sim; ++s) {
    std::vector<double> result = op->simCond(conditioning_data);
    // store those results somewhere...
}
```

Running that code either crashes or produces garbage (NaN or otherwise entirely wrong values). At the moment, what works is:
```C++
Model* model = ...;
MeshETurbo* mesh = ...;

MatrixSquareSymmetricSim* invnoise = ...;

std::vector<double> conditioning_data = ...;

#pragma omp parallel for
for (int s = 0; s < nb_sim; ++s) {
    // All those objects have to be created again for each thread!
#pragma omp critical // two PrecisionOpMulti cannot be created at the same time in different threads!
    PrecisionOpMulti* precision = new PrecisionOpMulti(model, { mesh }, true /* or false i.e., stencils or not */);

    ProjMulti* proj = ...; // maybe that also should be in a critical section?

    SPDEOp* op = new SPDEOp(precision, proj, invnoise, precision, proj); // that one can be outside of critical (i.e. truly parallel)

    std::vector<double> result = op->simCond(conditioning_data);
    // store those results somewhere...
}
```

There are several issues here. First is the fact that all this isn't documented and I found out by trying (but with parallel code, this means I cannot be sure that my last code actually works, as opposed to just a stroke of luck that it currently does but may break with e.g., more CPUs or a different scheduling!).

[...]

PR #507, opened 2025-04-15
Add convenience functions in SPDEOp to access some members

remi-moyen:
I need this to tweak some of `SPDEOp` behaviour, in my case to reimplement `simCond()` to work around the issues with the RNG not being properly reproducible/thread-safe (see #495 and #496).

### DIFF
--- a/include/LinearOp/SPDEOp.hpp
+++ b/include/LinearOp/SPDEOp.hpp
@@ -55,9 +55,11 @@ public:
                                   const MatrixDense& drifts) const;
   VectorDouble simCond(const VectorDouble& dat) const;
 
-  const ASimulable* invNoise() const { return _invNoise; }
-  const PrecisionOpMulti* Qsimu() const { return _QSimu; }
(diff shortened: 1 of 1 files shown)


NDesassis (changes requested):
Could we use the syntax with get? Example: getInvNoise()

NDesassis:
> Good point, I didn't really think about any syntax, sorry...
> 
> I will also add the other 2 missing functions (for `popkrig` and `proj`), for completeness.
> 
> The names of all those are not very consistent, what do you prefer?
> 
>     * getPop() or getPopKrig() ?
> 
>     * getPop() / getPopSimu() or getPopKrig() / getPopSimu() ?
> 
>     * getPop() or getQ() ? (and getPopSimu() or getQSimu() ?)
> 
> 
> And `proj`, getProj() or getProjKrig()?
> 
> Or something else entirely...

Generally, if we have an object _name, we simply use getName().

NDesassis (approved):
Thx Remi

PR #510, opened 2025-04-18
include RankHandler in Kriging

NDesassis:
The aim of this PR is to use RankHandler in the Kriging computation in order to implement multivariate in the optimized version of the Kriging.


fabien-ors (approved):
Thank you Nicolas

PR #511, opened 2025-04-22
Pluging inverse of cholesky for sparse matrix

drenard77810:
In this branch, the CholeskySparse class has been incremented by the possibility of calculating the inverse of the Cholesky decomposition performed on a Sparse Matrix (using the method 'partial_inverse' of the new class CholeskySparseInv).
Note: This algorithm assumes that the decomposition has been performed in LLT.
However, in its current version, CholeskySparse performs a LDLT decomposition and stores the relevant matrix in its member '_factor'.
This member CANNOT be used directly by 'partial_inverse'. Therefore a second decomposition (using LLT method this time) is performed locally ... simply for the sake of using 'partial_inverse' method as is.

In the future:
- either one can modify ALL the usages of _factor in the class CholeskySparse (as the result of a LLT rather than a LDLT)
- or we consider that the loss of time linked to this second decomposition is not important due to the very limited usage of this function (only used for calculating the variance of estimation error in SPDE for the time being). 
-

### DIFF
--- a/include/LinearOp/ACholesky.hpp
+++ b/include/LinearOp/ACholesky.hpp
@@ -36,7 +36,6 @@ public:
   int solveMatrix(const MatrixDense& b, MatrixDense& x) const;
   bool isReady() const { return _ready; }
 
-
   VectorDouble invLtX(const VectorDouble& vecin) const;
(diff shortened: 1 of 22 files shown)


fabien-ors (approved):
Thx Didier

PR #513, opened 2025-04-23
Remove roxygen2 from nonreg workflows, disable BUILD_DOC in dev Makefile

pierre-guillou:
Following #494, this PR removes the installation of roxygen2 in the nonreg workflows. The workflows should run about 30 seconds faster now…

Also, the `BUILD_DOC` variable in the main Makefile has been switched to `OFF`. This means that `python_doc` and `r_doc` won't run except if the variable is explicitly set.

Enjoy,
Pierre


fabien-ors (approved):
Nice! Thx Pierre

PR #516, opened 2025-04-25
Improving plot.r

drenard77810:
In this branch, we have updated (and corrected) the graphics of plot.R (R version of plots using ggplot and gstlearn).
The main modifications are as follows:
- function ggplot() should replaced by plot.init() which offers some additional possibilities to specify parametes such as 'asp', 'xlim'...
- function ggPrintR() (which does not exist anymore) should be replaced by plot.end()
- function plot.grid() does not exist anymore. It is replaced by more specific functions such as plot.raster() or plot.contour(). The benefit is that there arguments become more generic ('name', 'flagLegend' or 'legendName') rather than ('nameRaster' or 'nameContour', 'flagLegdnRaster' or 'flagLegendColour')
- similarly, plot.point() does not exsit anymore. It is replaced by more specific functions such as plot.symbol() or plot.literal().

The old functions ggDefault() or ggDefaultGeographic() where used to manage (internally) the default values for geographic subsequent representations. They were difficult to manipulate. This internal storage of default values has been abandonned. As a consequence, the user must define the dimensions of the graphics in ALL the plots, using plot.init() arguments.


fabien-ors (changes requested):
Could you update CHANGES file by adding the following sentence please?

`R plotting feature refactoring: see https://github.com/gstlearn/gstlearn/pull/516 for details.`

fabien-ors (changes requested):
Another request...
Previously, the name of the variable was used as the default plot title. Now, there is no default title displayed.
For example, in the first plot of [Multivariate Geostatistics.pdf](https://github.com/user-attachments/files/19906690/Multivariate.Geostatistics.pdf) (generated by your new plot.R interface) compared to the same course in the previous release: https://soft.mines-paristech.fr/gstlearn/1.6.0/courses/r/06a_Multivariate.html
Could it be possible to display a default title?

drenard77810:
I read your comment. I am not sure that this feature is completely elaborated for the following reasons:
1) I can pretend that it is the responsibility of the user to add a title (using plot.decoration)
2) If plot.raster() (alias to by plot performed on a DbGrid) produces a title built from the variable name, it may overwrite the title that the user may have explicitly defined using a call to plot.decoration (placed beforehand in the same sequence)
3) If a sequence is composed of:
plot.init() +
plot.raster(dbgrid, name="toto") +
plot.contour(dbgrid, name="titi") + 
plot.end(p)

Which variable should be promotted as title?
Thanks for providing me with a clever algorithm for the latter case.

fabien-ors:
You are right.
I have no better idea.
plot.decoration should be called each time the user wants to display a title.

fabien-ors (approved):
Nice Didier!
Thank you

PR #518, opened 2025-04-27
Improve retro compatiblity with old linux distribution for R package

fabien-ors:
This branch aim at producing R packages which are compatible with old linux distributions.
The R publish workflow now uses an almalinux-8 docker generated here :
https://github.com/gstlearn/docker-generate/actions/workflows/create_almalinux-8_gstlearn-r.yml

Now, future versions of gstlearn packages (R or python) should be compatible with Centos 8+, RedHat 8+, Almalinux 8+, Fedora Core 16+, Ubuntu 16+ etc...


drenard77810 (approved):

NDesassis (approved):
Thx!

PR #527, opened 2025-05-05
Fix Python Swig memory leak 

pierre-guillou:
This PR fixes memory leaks when converting Numpy arrays to C++ `VectorDouble` in the Swig wrapper.

According to https://stackoverflow.com/questions/57783652/why-does-pysequence-getitem-return-a-new-reference, the function `PySequence_GetItem` returns an owning pointer. Using `Py_DECREF` releases the corresponding memory.

Pierre


fabien-ors (approved):
Thx Pierre !

PR #529, opened 2025-05-07
Chantier spde

drenard77810:
This constitutes the first step for the renewal of the API for SPDE applications.
It lacks the following improvements:
- make Mesh and Proj optional arguments in kriging and simulation API
- provide an interface for LogLikelihood.

Then it will be possible to clean:
- the old API interfaces for SPDE
- the obsolete classes (only used in these old interfaces)


NDesassis (approved):
Thx a lot Didier!

PR #535, opened 2025-05-19
Fix publish workflow due to C++20 std::map::contains

fabien-ors:
The function std::map::contains used in PrecisionOp is available with C++ 20+ only. Thus the publish workflow which runs on an old Debian distribution (C++17) fails. `contains` function has been replaced by the classical `find`.


pierre-guillou (approved):
lgtm

PR #539, opened 2025-05-19
Fix Swig wrapping of HDF5 methods 

pierre-guillou:
This PR fixes the missing HDF5 methods in the Python and R wrappers. Swig was missing the `-DHDF5` compilation flag when analyzing the header files.

A non-regression test has been added (in Python only).

Pierre


NDesassis (approved):
Thx for reactivity!

PR #556, opened 2025-05-25
Checking use of boost on github machines

drenard77810:
This branch is meant to write cross_platform functions for file manipulation.
In the first tile, we simply try to see if the code written using BOOST could function on the different platforms tested on Github.


drenard77810:
This PR is abandoned as it is out of my reach.

PR #557, opened 2025-05-26
Loglikelihood par Vecchia

drenard77810:
Adding the calculation of LogLikelihood using Vecchia approximation


NDesassis (approved):
Thx a lot. I merge.

PR #568, opened 2025-06-04
Don't call ASerializable::buildFileName in csv_table_read

### LINKED ISSUE #567: Cannot create a Db from a CSV file anymore
Since this PR: https://github.com/gstlearn/gstlearn/pull/564
The function Db::createFromCSV doesn't work anymore when the filename is **relative** and corresponds to a file in the current directory.
We can see the problem by doing this:
```
import gstlearn as gl
import urllib.request
filename,head = urllib.request.urlretrieve("https://soft.mines-paristech.fr/gstlearn/data/Scotland/Scotland_Elevations.csv", "download_file.csv")
db = gl.Db.createFromCSV(filename)

```
Reason: The ASerializable::buildFileName function doesn't behave like before. When status is 1 (read a file), the function now adds "gstlearn_dir" folder in the path of the file name.

pierre-guillou:
Avoid calling `ASerializable::buildFileName` in `csv_table_read()`. No need to handle containers and prefixes when loading a CSV file.

Fix #567.

Pierre

### DIFF
--- a/src/Core/convert.cpp
+++ b/src/Core/convert.cpp
@@ -532,10 +532,9 @@ int csv_table_read(const String &filename,
   String na_string = csvfmt.getNaString();
 
   String line;
-  String filepath = ASerializable::buildFileName(1, filename, true);
   // Open new stream
   std::ifstream file;
-  file.open(filepath, std::ios::in);
+  file.open(filename, std::ios::in);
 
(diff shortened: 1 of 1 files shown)


fabien-ors (approved):
Thx Pierre

PR #575, opened 2025-06-06
Add python test dependencies in uv environment

fabien-ors:
This should fix check_py and check_ipynb on our computers


pierre-guillou (approved):
Works for me, with and without activating a venv.

We may also want to clean some redundant `uv pip install` statements in the nonreg workflow.

PR #578, opened 2025-06-09
Optim new3

drenard77810:
This branch is to conclude the new Fitting.
The branch should not be closed afterwards as lot of work is left to be done ... before cleaning


fabien-ors (approved):
Nice job !

PR #579, opened 2025-06-10
Correction bug n bez

drenard77810:
This branch concludes the correction of a bug found in the calculation of Global Estimate par N. Bez.
The error was located in the calculation of the variance of estimation error (which was coming out as constantly equal to 0).
Very improbable indeed.
This is now fixed (see non-regression test bench_GlobalKriging.cpp and bench_Krige.cpp


fabien-ors (approved):
Thx Didier

PR #582, opened 2025-06-11
Doc macro

NDesassis:
The aim of this PR is to generate documentation (for doxygen) when the macros FORWARD_METHOD and FORWARD_METHOD_NON_CONST are used.

First, the macros are not documented anymore by doxygen (thanks to DOXYGEN_EXCLUDE_SYMBOLS flag in CMake).
 Then a python script (macrodoc.py) is called when CMake creates the doxygen documentation.
This scripts searchs the FORWARD_METHOD calls, identify the method name which is plugged and the class type this plug refers to. Then it searches in the folder of this class the signature of the method and generate a specific documentation. If the method is not found in the class, the full inheritance hierarchy is explored.

Other modifications:
Note that all the python scripts called by CMake have been grouped in a new folder tools/scripts.
macrogenerate.py used to create some files in the CMAKE_SOURCE_DIR. Now the files are created in CMAKE_BINARY_DIR which let the source repository clean.


NDesassis:
Suggestions of copilot have been taken into account.

drenard77810 (approved):

PR #586, opened 2025-06-17
Fix publish due to reference_wrapper for R under windows 2022

fabien-ors:
The include file <functional> in ListParams.hpp is mandatory for the platform.


pierre-guillou (approved):
Seems good

PR #589, opened 2025-06-18
Add display in test_Basic

fabien-ors:
Add some display in test_Basic so that all 3 tests (cpp, py and R) generates exactly the same outputs.

### DIFF
--- a/tests/py/output/test_Basic.ref
+++ b/tests/py/output/test_Basic.ref
@@ -1,0 +1,195 @@
+
+Model characteristics
+=====================
+Space dimension              = 2
+Number of variable(s)        = 1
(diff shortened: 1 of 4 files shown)


fabien-ors:
> While the outputs of the R & Python tests are quasi-identical (except some "NULL" in the R output), the C++ result file contains different values. Is it expected ?

Nice point Pierre.
I assume that it is due to the random numbers generation. We could take a look to that later.
Regarding the NULL keyword in R output, this is a known issue. It can be hidden by encapsuling calls to 'display' method with the function 'invivsible'. But the objective of these Basic scripts is to maintain alive 3 scripts having the same workflow (as a demonstrator)

PR #597, opened 2025-07-01
Suppress mem alloc

drenard77810:
This branch is dedicated to reducing (why not suppressing) the old-fashion core allocation using mem_alloc() and mem_free().
Still some work left but ht number of modified files is huge: it is godo to push this partial branch... just to check on Github machines.


fabien-ors (approved):
Thx Didier

PR #603, opened 2025-07-04
ASerializable: Remove creation of "gstlearn_dir" folder

### LINKED ISSUE #570: Improve serialization source/destination folder management
Currently in gstlearn (1.7.2), the input/output folder for reading/writing Neutral Files (relative paths only) is:
- the one pointed by the environment variable GSTLEARN_OUTPUT_DIR (if defined) - or
- the folder named "gstlearn_dir" in the current directory (otherwise)

That means that if you want to create a gstlearn object from a neutral file using `createFromNF` with a relative path, you must put the file in a specific directory (not the current directory). This is not convenient.

The HDF5 serialization doesn't follow the same logic (only read/write files from/to current directory).

The folder management for serialization should be made more convenient.

pierre-guillou:
Without the `GSTLEARN_OUTPUT_DIR` env var defined, read & write files in the current working directory.

Fix #570.

Pierre

### DIFF
--- a/src/Basic/ASerializable.cpp
+++ b/src/Basic/ASerializable.cpp
@@ -183,8 +183,8 @@ String ASerializable::buildFileName(int status, const String& filename, bool ens
   fileLocal.clear();
 
   // container name: first search for the GSTLEARN_OUTPUT_DIR
-  // environment variable, then if empty create a `gstlearn_dir'
-  // folder in the current directory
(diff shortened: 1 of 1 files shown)


fabien-ors (approved):

fabien-ors (approved):
Thx Pierre

PR #607, opened 2025-07-08
Cleaning memalloc

drenard77810:
This branch concludes a long run of standard modifications performed on lot of files, including:
- suppressing some mem_alloc() and mem_free() calls (turned to more modern statement new /delete)
- replacing #include <math.h> into #include <cmath> (which is recommanded by clang-tidy)
- similar for other includes (searching for occurences of XXX.h>


fabien-ors (approved):
Nice cleanup! Thank you!

PR #609, opened 2025-07-10
Correction for Global

drenard77810:
In this branch:
- we continue chasing the old mem_alloc() and mem_free() statements. Work is still in progress

The branch has also served:
- to check the read/write operation for neutral files (in ASCII or HDF5 formats)
- to exhibit the weights for Global estimation: there were not printed when using the Kriged solution

An extra test has been implemented (in PY) to test that the Global estimates (using Arithmetic Mean and Kriging) are correct ... as well as the contents of the output structure (Global_Results)


fabien-ors (approved):
Thx Didier

PR #614, opened 2025-07-16
Vecchia revisited

drenard77810:
This branch concludes the ability of Vecchia to process a multivariate case.
The test test_Cecchia.cpp contains a new 'mode' allowing the test of Kriging with a 100 data set (2 variables) on a grid of 100x100


fabien-ors (approved):
Thx Didier

PR #627, opened 2025-08-05
Add back macro GSTLEARN_VERSION_NUMBER

### LINKED ISSUE #626: Restore GSTLEARN_VERSION_NUMBER macro
This macro was removed in commit https://github.com/gstlearn/gstlearn/commit/f8b0b11253e239c554bf871fe365542cc2c477dc which added pre-release tags.

I don't see a reason to remove the macro. While it does not contain the pre-release tags information, it is still useful to keep code compatible with more than one version of gstlearn at the same time (and I don't think the missing pre-release tags is an issue, code that will want to be compatible with several versions is long-term code that should not use pre-release tags anyway, or that can just use the version number of the future release (without the tag) if they really need to).

remi-moyen:


fabien-ors (approved):
Ok for restoring GSTLEARN_VERSION_NUMBER

PR #629, opened 2025-08-05
Fix compilation warning (overflow of constant)

### LINKED ISSUE #628: Compilation warning on constant overflow
On Windows, the following line causes a compilation warning about a constant overflowing:
https://github.com/gstlearn/gstlearn/blob/8edf78ac510765730c1d1ed8157d646db75d0a5e/include/Covariances/CovAniso.hpp#L184

This is because `getMinOrder()` returns an `int` but `TEST` is a `double` of value 1e30 which overflows when stored in an `int`. I assume the intent was to use `ITEST` here and this is just a copy-paste error.

remi-moyen:


NDesassis (approved):
Thx Remi.

PR #635, opened 2025-08-07
Replace mem_alloc & malloc with std::vector

pierre-guillou:
This PR replaces a few `mem_alloc`s & `malloc`s with `VectorInt`, `VectorDouble` or `std::vector` when applicable.

There is still a few `mem_alloc` left, in particular in `krige.cpp` and `spde.cpp`. Those ones seems harder to tackle.

Pierre

### DIFF
--- a/include/LinearOp/CholeskySparseInv.hpp
+++ b/include/LinearOp/CholeskySparseInv.hpp
@@ -25,9 +25,9 @@ template<typename SpMat>
 class MatchPattern
 {
   using T = typename SpMat::value_type;
-  StorageIndex* m_outer;
-  StorageIndex* m_inner;
-  T* m_val;
+  std::vector<StorageIndex> m_outer;
+  std::vector<StorageIndex> m_inner;
+  std::vector<T> m_val;
   StorageIndex m_cols;
   StorageIndex m_nnz;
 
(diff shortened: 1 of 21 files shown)


include/LinearOp/CholeskySparseInv.hpp
```diff
@@ -114,7 +107,7 @@ typename SpChol::MatrixType partial_inverse(
   const SpMat& L     = llt.matrixL();
   SpMat Qinv         = L.template selfadjointView<Eigen::Lower>();
 
-  for (Id i = ncols - 1; i >= 0; --i)
+  for (int i = ncols - 1; i >= 0; --i)
```
drenard77810, line 110: Are you sure you want to transform a Id int int

src/Core/util.cpp
```diff
@@ -971,23 +902,20 @@ void print_last_message(void)
...
+VectorInt ut_split_into_two(Id ncolor, Id flag_half, Id verbose, Id* nposs)
 {
   Id p, nmax, ncomb, np, lec;
-  Id* mattab;
+  VectorInt mattab;
 
   /* Initializations */
 
-  p      = (flag_half) ? static_cast<Id>(floor((double)ncolor / 2.)) : ncolor - 1;
+  p      = (flag_half) ? static_cast<int>(floor((double)ncolor / 2.)) : ncolor - 1;
```
drenard77810, line 912: transformer en Id?

drenard77810 (approved):

PR #644, opened 2025-08-12
Covariance gradient

drenard77810:
This is the first implementation of the new CovGradient class.
A non-regression cpp test (test_Gradient.cpp) has been developed to check that we get the same results with the new class as with the old test where the Depth-Gradient case was emulated using a funny LMC model (this case corresponds to the Jeu8 data set of the non-regression Data tests).

In the current version, the covariance between Z and Gradient, or along Gradients, are performed using the Numerical integration. In the future, we should try to use a formal derivation (defined for each covariance) rather that with this numerical  solution.

Moreover the current version is limited to the monovariate case in 1-D or 2-D.


fabien-ors (approved):
Thx Didier

PR #659, opened 2025-08-28
Replace std::filesystem by boost::filesystem (and prevent R crash under Linux)

fabien-ors:
The use of std::filesystem (for serialization) in R packages under Linux was not operational. Calling dumpToNF or createFromNF was crashing R. It has been replaced by boost::filesystem.

Details: 
Linux R packages are built using an AlmaLinux 8 docker (old GCC < 9) to ensure maximum backward compatibility with old Linux distros. Thus, R packages delivered to our CRAN are not compatible with recent Linux distros which have a different implementation of std::filesystem. Using boost::filesystem (older) fix the issue.

Coverage tests have been upgraded to test serialization on all platforms.


fabien-ors:
No more need of test-dumpToNF-under-R branch since #662

PR #660, opened 2025-08-29
Nouvelle classe potential

drenard77810:
This chunk has been performed in order to:
- develop a new class API/Potential which also brings 3 free methods which are:
  . krigingPotential(), simulatePotential() and xvalidPotential()
- suppress the old code from Core/potential
- adapt the cpp and ipynb tests


fabien-ors (approved):
Thx Didier !

PR #673, opened 2025-09-10
Fix a Windows installation bug (missing .lib)

### LINKED ISSUE #672: Incomplete installation on Windows
When installing the shared library on Windows, the `.lib` file (needed to compile another project with gstlearn) is not longer installed.

A recent change (by myself!) modified the `install()` part of `install.cmake` (to handle installing the static library). The relevant part is:
https://github.com/gstlearn/gstlearn/blob/dbd4c40e4187159306cac8c1d70b519e7dafd65c/cmake/install.cmake#L52-L58

Note how for the `ARCHIVE` component (i.e., the one for the `.lib` file) there is no `DESTINATION`. [CMake doc](https://cmake.org/cmake/help/latest/command/install.html#targets) says:

> For shared libraries on DLL platforms, if neither RUNTIME nor ARCHIVE destinations are specified, both the RUNTIME and ARCHIVE components are installed to their default destinations. If either a RUNTIME or ARCHIVE destination is specified, the component is installed to that destination, and the other component is not installed. If both RUNTIME and ARCHIVE destinations are specified, then both components are installed to their respective destinations.

I don't understand why CMake works this way, but this means that *on Windows* (but not on other platforms!), the DESTINATION **must** be specified for both components, we can't just specify the component itself.

remi-moyen:


fabien-ors (approved):
Thx Rémi !

PR #685, opened 2025-09-18
Snapshot of v1.8.2

fabien-ors:
This snapshot has an additional commit compared to tag stable_v1.8.2.
This commit aim at fixing the CI for demos/courses.
It does not impact the v1.8.2 gstlearn library/packages behavior.


pierre-guillou (approved):
Everything seems ok here.

PR #687, opened 2025-09-19
GitHub Actions: Enable clang-check based parsing of source & header files

pierre-guillou:
This PR enables the `check-warnings` job of the `check_code` workflow to pass the `clang-check` tool on gstlearn source & header files to ensure they are parsed without warnings or errors.

This is done on a Cartesian product of several configurations: Debug vs. Release, HDF5 or not, Boost vs. `std::span`. More configurations can be added, up to a maximum of 256 parallel jobs.

Performing `clang-check` also on header files ensure that they don't depend on each other or on the include order.
Two missing includes were found and fixed.

Finally, a CMake option has been introduced to expose `USE_BOOST_SPAN` to the user.

Pierre

### DIFF
--- a/.github/workflows/check_code.yml
+++ b/.github/workflows/check_code.yml
@@ -11,11 +11,6 @@ on:
         required: false
         default: ''
 
-concurrency:
-  # allow a failing job inside a matrix to stop the other jobs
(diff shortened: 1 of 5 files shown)


fabien-ors (approved):
Nice job !

PR #717, opened 2025-10-06
Correct lower and upper in gibbs

drenard77810:
When running gibbs_sampler facility while defining simultaneously Lower bound (Loc::L), Upper bound (Loc::U) and exact value (Loc::Z), a severe conflict may rise as the array dimensioning was defined by the number of active and defined samples for the Z-locator variable. This remains true in the multivariate case.
Moreover, the algorithm was not clear to decide if the sample was hard or soft when the three variables were defined.
Now things are much simpler...

As a prior step and if the three variables are defined (for a given sample and for any variable):
- the value of the Z-locator variable is copied into the corresponding L-locator and U-locator samples (regardless of there previous contents)
- the Z-locator is cancelled, so that the rest of the algorithm is performed only using Lower and Upper bound (which must still be consistent)

### DIFF
--- a/src/Core/simtub.cpp
+++ b/src/Core/simtub.cpp
@@ -556,7 +556,7 @@ static void st_check_facies_data2grid(Db* dbin,
     for (isimu = 0; isimu < nbsimu; isimu++)
     {
       facres = static_cast<Id>(dbgrid->getSimvar(ELoc::FACIES, jech, isimu, 0, ipgs,
-                                     nbsimu, 1));
+                                                 nbsimu, 1));
       if (flag_show)
       {
         if (facdat == facres)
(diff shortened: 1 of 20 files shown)


src/Core/simtub.cpp
```diff
@@ -1505,99 +1505,148 @@ Id gibbs_sampler(Db* dbin,
...
-    }
-  }
-
   /* Model */
 
   if (model == nullptr)
   {
     messerr("No Model is provided");
-    goto label_end;
+    return 1;
```
fabien-ors, line 1531: There is no more call to proportion_manage after the label_end.
Is it the required behavior ?
(same remark below)

PR #718, opened 2025-10-06
Robustifiy Kriging (no variable defined)

drenard77810:
In this branch, we set an additional test which check that, for Interpolators, the number of variables defined in the model matches the count of variables defined in the Input Db (if defined: ... pay attention to non conditional simulations where no Input Db is defined).


fabien-ors (approved):
Nice Didier !

PR #730, opened 2025-10-17
Enable & fix -Woverloaded-virtual on Unix platforms

pierre-guillou:
This PR enables the `-Woverloaded-virtual` GCC & Clang warning, which prevents two virtual member functions to overload one another (i.e. same name but different parameters or cv-qualifiers).

A few methods were renamed to fix the corresponding warnings.

Pierre


pierre-guillou:
NB: Random (?) fail on Tuto_PCA_and_MAF_rmd_cmp

fabien-ors (approved):
Thx Pierre !

PR #731, opened 2025-10-20
GitHub Actions: Fix warnings in publish workflows

pierre-guillou:
This PR enables `-Werror` when building `gstlearn` in the publish workflows (except `publish_r_linux` which currently depends on an outdated SWIG in the Docker container).

A few warnings were fixed in the C++ source code.

Pierre


fabien-ors (approved):
Thx Pierre

PR #733, opened 2025-10-21
Improve db  create from polygon

drenard77810:
This branch corresponds to a suggestion from N. Bez to improve DbGrid::createFromPolygon:
- allow the arguments 'nodes' and 'dcell' to remain undefined by the user. The rule is the following (as explained in the documentation of the function):
- If 'nodes' is undefined, it is defaulted to 10 for each space dimension
- If 'nodes' is defined, 'dcell' is derived so that the non-rotated grid covers the whole polygon extension
- if 'dcell' is defined, 'nodes' is derives so that the non-rotated grid covers the whole polygon extension
- If both are defined, 'dcell' prevails

The argument 'nodes' has been renamed into 'nx' and 'dcell' into 'dx'.

An additional argument 'addSelection' allows adding the selection of the grid nodes belonging to the polygon in the newly created grid.

An extra bug has been fixed in the calculation of the Polygon extension.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,8 @@ This file contains the logs between consecutive releases
 
 Release 1.10.0
 ==============
+- Improvement in DbGrid::createFromPolygon (better management of default value + ability to define the selection in the newly create Grid File)
+- Correct the Polygon extension calculation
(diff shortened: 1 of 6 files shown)


fabien-ors (approved):
Very nice PR!
Thx Didier!

PR #758, opened 2025-11-03
Correction of bug 745

### LINKED ISSUE #745: Fix R plots size when using ggarange
When using only one legend for two plots in the same ggarrange call, the plots have not the same size anymore (even with asp=1)
The following R code...
```
suppressWarnings(suppressMessages(library(gstlearn)))

mygrid = DbGrid_create(nx=c(100,150), dx=c(1.,1.))
mymodel = Model_createFromParam(ECov_SPHERICAL(), sill = 1, ranges = c(50, 30), angles = c(30, 0))
simtub(NULL, mygrid, mymodel, nbsimu=2, nbtuba=1000, namconv=NamingConvention("Simu"), seed=14341)

p1 = plot.init(asp=1)
p1 = p1 + plot.raster(mygrid, name="Simu.1", limits=c(-3.5, 3.5))
p1 = p1 + plot.decoration(title="Simulation 1", xlab="Easting", ylab="Northing")
p2 = plot.init(asp=1)
p2 = p2 + plot.raster(mygrid, name="Simu.2", flagLegend=TRUE, legendName="", limits=c(-3.5, 3.5))
p2 = p2 + plot.decoration(title="Simulation 2", xlab="Easting", ylab="")
p = ggarrange(p1, p2, ncol = 2, nrow = 1)
plot.end(p)
```
produces the following plot:

Image

drenard77810:
Fix #745 

This branch check the reason of a strange dimensionning when cobining severalplots in the same figure, while producing a legend.
According to chatGPT, it is simply a matter of the way to call ggarrange() (which arranges the figures in the same plot).

In order to understand the conclusion, the tutorial 03_Graphic.Rmd has been amended with a specific example of this sort.

### DIFF
--- a/doc/courses/r/03_Graphics.Rmd
+++ b/doc/courses/r/03_Graphics.Rmd
@@ -2,11 +2,13 @@
 title: "Tutorial on Graphics"
 author: "gstlearn Team"
 output:
+  pdf_document:
+    toc: true
   html_document:
     df_print: paged
     toc: true
-  pdf_document:
-    toc: true
+editor_options: 
+  chunk_output_type: console
 ---
 
 <!-- SUMMARY: Use the library gtlearn.plot (based on matplotlib) to visualize all the objects of gstlearn library   -->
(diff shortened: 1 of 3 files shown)


tests/r/test_Basic.R
fabien-ors: Does this modification really needed?
Note that the test_Basic in the up-to-date dev branch no more generates plots.
I suggest you to rebase from dev (non-reg has been fixed)

PR #760, opened 2025-11-03
Add python 3.14 compatibility and upgrade to R 4.5.2 & fix projections and polygons

fabien-ors:
This PR adds the compatibility to python 3.14 and R 4.5.2.
Doing so, the dependency to `geopandas` (which was used for python coordinates projections) has been replaced by `pyproj`. `geopandas` and `shapely` are no more needed by `gstlearn.plot` python module.
It is now possible to overlay polygon labels using `ax.polygon` python function.
The `proj` module has been simplified and a new tutorial has been added (Tuto_ProjectionHalieutic.ipynb) to demonstrate how to use it. In the future the `proj` module will be improved to facilitate the coordinates transformation of a `Db` or a `Polygon` object.
The method `PolyElem::getSurface` has been fixed (the surface calculation of a polygon was wrong).

### DIFF
--- a/src/Polygon/PolyElem.cpp
+++ b/src/Polygon/PolyElem.cpp
@@ -71,15 +71,15 @@ String PolyElem::toString(const AStringFormat* strfmt) const
   return sstr.str();
 }
 
-void PolyElem::getExtension(double *xmin,
-                           double *xmax,
-                           double *ymin,
-                           double *ymax) const
+void PolyElem::getExtension(double &xmin,
+                            double &xmax,
+                            double &ymin,
+                            double &ymax) const
 {
-  *xmin = getXmin();
-  *ymin = getYmin();
-  *xmax = getXmax();
-  *ymax = getYmax();
+  xmin = getXmin();
+  ymin = getYmin();
+  xmax = getXmax();
+  ymax = getYmax();
 }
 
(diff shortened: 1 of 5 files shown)


src/Polygon/PolyElem.cpp
```diff
@@ -259,6 +259,52 @@ bool PolyElem::inside3D(double zz) const
...
+
+  if (ABS(area) < EPSILON8)
+  {
+    // Degenerated polygon
+    return {0., 0.};
+  }
+
+  area *= 0.5;
+  centroid[0] /= (6. * area);
+  centroid[1] /= (6. * area);
```
pierre-guillou: Where does this `6.` magic number come from?

src/Polygon/Polygons.cpp
```diff
@@ -344,6 +344,13 @@ void Polygons::getExtension(double* xmin,
   }
 }
 
+VectorDouble Polygons::getExtensionAsVD() const
+{
+  VectorDouble ext(4);
+  getExtension(&ext.data()[0], &ext.data()[1], &ext.data()[2], &ext.data()[3]);
```
pierre-guillou: `.data()` seems unnecessary

pierre-guillou (approved):

PR #767, opened 2025-11-07
Basculement de qqs commandes de VectorHelper vers VectorNumT

drenard77810:
This branch is meant to clean the contents of VectorHelper as much as possible and to move the functions in VectorNumT using templates. This should reduce the number of functions (not repeated anymore for VectorDouble and VectorInt).

Some more functions should be moved gradually.

A special trick should even enabme to consider the VectorVectorInt and VectorVectirDouble in the same operation (to be demonstrated).

### DIFF
--- a/tests/py/output/test_Template.ref
+++ b/tests/py/output/test_Template.ref
@@ -1,12 +1,12 @@
 rtab1
 [2.1 3.2 4.1 NA 8.3]
 Vector of < double > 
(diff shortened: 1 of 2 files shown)


fabien-ors (approved):
Thx Didier!
Honestly, I haven't reviewed all 206 modified files.

fabien-ors (approved):
Now it's nice :)

PR #770, opened 2025-11-17
Replace Vector copies with const references

pierre-guillou:
This PR replaces a few `VectorDouble` copies from methods returning references into full `const` references.

This should reduce the number of unwanted copies.

Pierre


NDesassis (approved):
Thx Pierre!

PR #772, opened 2025-11-18
Templating string

drenard77810:
This branch is meant to clean up the AStringable class. 
All methods dedicated to String manipulation are now moved into String.hpp and String.cpp (using lots of templated functionalities in order to avoid code duplication).

The next step will be to move all functions concerning message activity to a new class/file (already created but empty) simply named Basic/Message.cpp/hpp

This last operation is delayed as the number of files involved will be very large. As a matter of fact, in the include part, we will have to change include AStringable by include/Message in lots of files.

### DIFF
--- a/include/Basic/String.hpp
+++ b/include/Basic/String.hpp
@@ -156,7 +156,7 @@ GSTLEARN_EXPORT String toStrVectorVec(const String& title,
 /**
  * @brief Converting the contents of a String into double or integer
  *
- * @param string Input string
+ * @param s Input string
  * @param dec Number of decimals (only used for Double conversion)
  */
 // Template générique (non défini)


drenard77810:
Some final cleaning are required. This is the reason why this PR has been moved into DRAFT.

drenard77810:
As the number of files modified is large anyhow, I plan to:
- submit this draft PR to check that everything is OK in my template functions on all platforms
The next step consists in:
- renaming tab_printg, tab_printi, tab_prints into printElement (for overload)
- renaming print_vector into printVector
- renaming print_Matrix into printMatrix
- renaming print_trimat into printTriMat
- call one of these in the dump function within VectorNumT class (in order to replace the stuff developed specifically there)
Move all these printing functions in the new file (already created) called Message.hpp and Message.cpp. This movement will modify the include file from AStringable.hpp into the new Message.hpp.

This systematic modification was initially meant to be postponed by another branch. But as the number of files modified is already large, I decided to include this EARTHQUAKE in the current Draft PR.
The PR ultimately will be moved into an official one (rather than a Draft).

This will

drenard77810:
All the essential renaming activity has now been performed (except the dump function stored in VectorNumT which must still be arranged).
Next step will be big move (from AStringable to Message) which will lead to a general change of include files.

drenard77810:
The last move has been performed.
However...
In order not to destroy the include dependency, the class AStringable (which previously contained the messages) is simply including the Message.hpp.
In the future, we should REALLY include AStringable.hpp when needed, and not in order to include Message.hpp indirectly. This is pretty simple as the only classes that NEED AStringable are the ones which derive from the AStringable class. All the others in fact, sgould include Message.hpp.

This PR is now an official one (not a Draft anymore).

PR #773, opened 2025-11-19
Fix all calls to getAnisoAngles function

fabien-ors:
The calls to getAnisoAngles were not totally fixed (no more argument in the function).
This has been fixed in marimo and widgets python features.


NDesassis (approved):
Perfect! I wonder if we shouldn't rename getAnisoAngle simply by getAngle ?

Issue #741, opened 2025-10-24
ProjMatrix is wrong for points exactly on mesh apices if there are missing elements in the mesh (selection)


remi-moyen:
Consider the following test where a `gstlrn::MeshETurbo` is built from a `gstlrn::DbGrid` with a selection such that some elements are missing (in this example I'm removing one element from the middle of the grid but I am almost certain the same issue would happen if removing elements from the border of the grid, which is a more common use-case).

Then a `gstlrn::ProjMatrix` is created on this grid and mesh. Since the mesh has exactly the same apices as the grid (there are some missing ones but they are missing in both objects), the matrix should be the identity matrix, but it is not.

```c++
gstlrn::DbGrid* db_out = gstlrn::DbGrid::create({ 5, 5 });
db_out->addColumns(
  {
    1., 1., 1., 1., 1.,
    1., 0., 1., 1., 1.,
    1., 1., 1., 1., 1.,
    1., 1., 1., 1., 1.,
    1., 1., 1., 1., 1.,
  }, "SEL", gstlrn::ELoc::SEL
);
gstlrn::MeshETurbo* mesh = gstlrn::MeshETurbo::createFromGrid(db_out);

gstlrn::ProjMatrix* proj2D = gstlrn::ProjMatrix::create(db_out, mesh, -1, true);
std::cout << proj2D->toString();
```
The output is:
```
Mesh Barycenter
===============
Sample    1 assigned to Grid Node    1 :    0 (1.00)    1 (0.00)    5 (0.00)
Sample    2 assigned to Grid Node    2 :Sample #2 does not belong to the meshing
Sample    3 assigned to Grid Node    3 :    2 (1.00)    3 (0.00)    6 (0.00)
Sample    4 assigned to Grid Node    4 :    3 (1.00)    4 (0.00)    7 (0.00)
Sample    5 assigned to Grid Node    5 :    3 (0.00)    4 (1.00)    7 (0.00)
Sample    6 assigned to Grid Node    6 :Sample #6 does not belong to the meshing
Sample    8 assigned to Grid Node    8 :    6 (1.00)    7 (0.00)   11 (0.00)
Sample    9 assigned to Grid Node    9 :    7 (1.00)    8 (0.00)   12 (0.00)
Sample   10 assigned to Grid Node   10 :    7 (0.00)    8 (1.00)   12 (0.00)
Sample   11 assigned to Grid Node   11 :    9 (1.00)   10 (0.00)   14 (0.00)
[...]

PR #776, opened 2025-11-20
ProjMatrix is wrong for points exactly on mesh apices if there are missing elements in the mesh (selection)

remi-moyen:
This implements the solution I suggested in #741.

I don't really like this code, I think it might be better to integrate that into `resetProjFromDb()` (and let `_addElementToTriplet()` handle only the simple case of testing the mesh elements of a single cell), but this is way easier to implement this way.

### DIFF
--- a/tests/ipynb/output/Tuto_SPDE_PGS.asciidoc
+++ b/tests/ipynb/output/Tuto_SPDE_PGS.asciidoc
@@ -85,42 +85,42 @@ Data Base Statistics
 4 - Name Prop.1 - Locator p1
  Nb of data          =       4747
  Nb of active values =        752
- Minimum value       =      0.000
+ Minimum value       =      0.014
  Maximum value       =      0.867
- Mean value          =      0.218
- Standard Deviation  =      0.162
- Variance            =      0.026
+ Mean value          =      0.267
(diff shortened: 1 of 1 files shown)


pierre-guillou:
The failing test reference is the file ./tests/ipynb/output/Tuto_SPDE_PGS.asciidoc. 

The failing job should make available archives of job log files, at the bottom of their Summary view. In this case: https://github.com/gstlearn/gstlearn/actions/runs/19534301849?pr=776, the archive `demos-courses-python-logs` should contain a new `Tuto_SPDE_PGS.asciidoc`.

Replace the old file by the up-to-date one, commit & push the result and the CI should pass!

Pierre

remi-moyen:
Many thanks @fabien-ors for the comparison. I apologise for taking so long to answer...

I would indeed appreciate if someone could check that my changes are correct because I am not absolutely sure the differences are the kind that we expect with this change. Maybe there is another place in the code that needs to be changed as well? In my own application of SPDE this does not seem to be the case but maybe I'm missing something.

Also, I think my commit should change the points of the mesh that are valid, so I think that the shape of the output should be a bit different in the end? or maybe not, as the map is showing the grid and not the mesh...

PR #777, opened 2025-11-21
Optimize test_Francky

pierre-guillou:
This PR reduces the number of heap allocations in `test_Francky`. Measured speedup is around 40% on my laptop.

The main gains were obtained by using optimized versions of matrix operators, either through the `noalias` Eigen call (when we are sure not to do the op in-place) or by providing a specific, no-Eigen implementation for inverting 2x2 matrices (thus avoiding a call to `Eigen::PivotLU`).

Also, some allocations have been factored with `thread_local` variables.

Here are the `heaptrack` figures:

|                         | before     | after   |
|-------------------------|------------|---------|
| Exec time (s)           | 3.6        | 2.1     |
| Heaptrack exec time (s) | 17.8       | 3.5     |
| #Allocations            | 48 000 000 | 525 000 |
| #Temp allocations       | 10 000 000 | 31 000  |
| Memory cons (MB)        | 89         | 89      |

Pierre

### DIFF
--- a/src/Basic/Tensor.cpp
+++ b/src/Basic/Tensor.cpp
@@ -274,7 +274,8 @@ void Tensor::_direct2ToInverse2()
   }
   else
   {
-    _tensorInverse2.eigenMat().noalias() = _tensorDirect2.eigenMat().inverse();
+    _tensorInverse2 = _tensorDirect2;
+    _tensorInverse2.invert();
   }
 }
 


src/Basic/Tensor.cpp
NDesassis: We tried to avoid Eigen code outside the matrix classes.

src/Matrix/MatrixSquare.cpp
NDesassis: maybe a invert3x3 would be great!

NDesassis (approved):
Thx!

Issue #779, opened 2025-11-25
Bug in DbGrid::createCoarse after deleting a column


NDesassis:
import gstlearn as gl
db = gl.DbGrid.create([100,100])
db["toto"] = db["x1"]
db.deleteColumn("toto")
db["toto2"] = db["x1"]
a = gl.DbGrid.createCoarse(db,[2,2])

PR #783, opened 2025-11-25
Issue 779

drenard77810:
issue #779 

Nicolas pointed out a procedure where the author made a confusion between:
- the column corresponding to a variable
- its UID
The difference is crucial: the programmer should favor the UID (which is a unique variable identifier which does not vary across time, even when other variables [defined BEFORE the target variable] are destroyed. 
Moreover, this is even more dangerous when the confusion was made in a hidden sub-procedure. That is the case when Db::createCoarse() function was ultimately calling MigrateByAttribute() where the list of variables (from dbin) to be migrated was provided as the list of their column ranks rather than the list of their UIDs.
Obviously a safer programming should favor using the Variable names (which remain constant , well defined). No confusion is then allowed as all variable names are different.

As a conclusion: working with a list of variables indexed either by their column ranks and UIDS should be discarded.


NDesassis (approved):
Thx Didier!

PR #785, opened 2025-11-26
New method in class Db

drenard77810:
This branch has been used to introduce a new function in the Db class: displayStats().
It allows providing a Table (decorated matrix format) containing some statistics (possibly chosen be the user) on some variables (possibly chosen by the user... otherwise all of them).
This is a better version of the traditional display() facility performed on an element of the class Db: this required using the DbStringFormat() facility in order to select specifically the FLAG_STATS option. This was considered as painful and will be better replaced by this new feature.

The output Table can easily be printed using a display() facility on an element of the Table class.


fabien-ors (approved):
Thx Didier

PR #788, opened 2025-11-26
Conversion in latex

drenard77810:
This branch is meant to introduce a converter to Latex.
This will ultimately be provided for Target Language R and Python (only Python has been initialized so far).

Moreover this feature will be applied to very specific class elements: for the time being only to Matrices ... avoiding sparse ones and Tables).

It is performed using toLatex which follows the same methodology as toTL.


fabien-ors (approved):
Thx Didier!

PR #790, opened 2025-11-29
Final modifs on Message.hpp

drenard77810:
This branch corresponds to the latest modification suggested by Fabien and concerning the order of the arguments in the methods of class Message. It involves a large number of files and has been placed in high priority to avoid interefering in the future.


fabien-ors (approved):
Nice Didier.
Thx!

PR #793, opened 2025-12-02
New class MatEigen

drenard77810:
A new class is born: MatEigen. It contains the mechanism to construct Eigen decomposition (or generalized Eigen Decomposition).
The corresponding members (vector of Eigen values and matrix of eigen vectors) have been removed from MatrixDense, MatrixSquare and MatrixSymmetric.

### DIFF
--- a/tests/cpp/test_a_template.cpp
+++ b/tests/cpp/test_a_template.cpp
@@ -12,7 +12,6 @@
 #include "geoslib_define.h"
 
 #include "Basic/ASerializable.hpp"
-#include "Matrix/MatrixSymmetric.hpp"
 
 using namespace gstlrn;
 
@@ -23,8 +22,4 @@ int main(int argc, char* argv[])
   sfn << gslBaseName(__FILE__) << ".out";
   StdoutRedirect sr(sfn.str(), argc, argv);
   ASerializable::setPrefixName("test_a_template-"); // Here set the test name
-
-  Id neq       = 5;
-  auto* matrix = MatrixSymmetric::createRandomDefinitePositive(neq);
-  matrix->display();
 }


include/geoslib_define.h
```diff
@@ -65,17 +65,17 @@ using I32 = int;
...
+  {                                                                                        \
+    std::cout << "Method 'toTL' Not implemented yet (missing dependencies?)" << std::endl; \
+  };                                                                    
 
 // Same for toLatex
 #define DECLARE_TOLATEX                                                      \
   inline void toLatex() const                                                \
   {                                                                          \
-    messerr("Method 'toLatex' Not implemented yet (missing dependencies?)"); \
+    std::cout << "Method 'toLatex' Not implemented yet (missing dependencies?)" << std::endl; \
```
pierre-guillou: `std::cerr`?

pierre-guillou (changes requested):

src/Matrix/EigenVectors.cpp
```diff
@@ -62,9 +62,9 @@ EigenVectors& EigenVectors::operator=(const EigenVectors& r)
     _ready        = r._ready;
     _eigenValues  = r._eigenValues;
     _eigenVectors = r._eigenVectors;
-    _mat          = r._mat;
-    _nrows        = r._nrows;
-    _ncols        = r._ncols;
+    // _mat          = r._mat; // _mat is a reference, cannot be assigned
```
pierre-guillou: I'm not sure if this `operator=` works correctly when the copy here is disabled.
I would remove altogether the whole `operator=` definition and let the compiler handle that itself (it should `delete` the method). Same for the copy constructor (rule of zero).

PR #808, opened 2025-12-31
Fix crash in MatrixSquare::invertOutOfPlace()

### LINKED ISSUE #807: Crash in MatrixSquare::invertOutOfPlace() for matrices larger than 3x3
A recent set of commits (see #787) introduced `MatrixSquare::invertOutOfPlace()` (as a way to optimise the very common 2x2 and 3x3 cases) but introduced a bug when the matrix is not 2x2 or 3x3, as the result matrix is not resized first. Therefore trying to write to it directly will fail, unless by luck it already was of the correct size:

https://github.com/gstlearn/gstlearn/blob/bebc02aff3cb8a7d84f07fcdb4700fcb93200e72/src/Matrix/MatrixSquare.cpp#L154

The fix is to first resize the matrix, as is done in both `invert2x2()` and `invert3x3()`:

https://github.com/gstlearn/gstlearn/blob/bebc02aff3cb8a7d84f07fcdb4700fcb93200e72/src/Matrix/MatrixSquare.cpp#L128

I think this bug was missed through a combination of two things:

First, in almost all places where it is used (there are only 3 overall! so 2 out of 3 is "almost all..."), the result matrix is already of the correct size:
https://github.com/gstlearn/gstlearn/blob/bebc02aff3cb8a7d84f07fcdb4700fcb93200e72/src/LinearOp/ShiftOpMatrix.cpp#L742
and
https://github.com/gstlearn/gstlearn/blob/bebc02aff3cb8a7d84f07fcdb4700fcb93200e72/src/Basic/Tensor.cpp#L267
(this is less obvious in the 2nd case but I think in practice this is the case -- or maybe the matrix is never more than 3x3 in practice, I'm not sure)

The 3rd use of this function is where I found the bug:
https://github.com/gstlearn/gstlearn/blob/bebc02aff3cb8a7d84f07fcdb4700fcb93200e72/src/Mesh/MeshETurbo.cpp#L851
But as long as the mesh used is 2D (or 1D), the matrix used here is at most 3x3 and thus would not hit the problem. I am using a 3D mesh, therefore with a 4x4 matrix here, which is why I hit the issue.

The reason I'm saying all this is that I'm guessing there are no tests in the test suite with a 3D mesh, it might be worth adding one? There are enough specific code for 3D that it would be good if some tests covered it?

remi-moyen:


pierre-guillou (approved):
Seems I forgot this `resize` call, sorry.
`resize` was already called in `invert2x2` and `invert3x3` but forgotten in the general case.
Good catch!

PR #814, opened 2026-01-11
Model serialization avec nostat

drenard77810:
This branch is meant to integrate the non-stationarity feature in the ASerialization of a Model.
The tests are completed OK. 
The ASerializable function serialize() and deserialise() have been moved to public.
The class CovParamId now aggregates a ParamId directly (classes are linked).

A test_a_template prooves that a slight bug has been left and must be solved before this branch is considered as finished. Therefore it is simplya  draft for the time being.


pierre-guillou (approved):
Thanks Didier!

PR #817, opened 2026-01-14
Fix macro around HDF5 methods

pierre-guillou:
The `getGroupFullPath` & `getGroupParents` expose HDF5 data structures. They need to be under `#ifdef HDF5` macro guards.

Pierre


fabien-ors (approved):
Thx Pierre!

PR #818, opened 2026-01-15
Fix memory leaks

remi-moyen:
This PR fixes two memory leaks.

The one in `ACov::operator=()` is pretty self-explanatory.

The other in `ProjMulti` may be discussed? From the commit message:

`ProjMultiMatrix` has a flag in the constructor to indicate that the destructor must delete all sub-matrices, but there is nothing similar in `ProjMulti` itself. So now the flag has been moved from `ProjMultiMatrix` to `ProjMulti` to fix this.

Note that, since there is one additional argument in constructor, if you were calling `new ProjMulti(..., true)` to set the `silent` flag, the same code will still compile but now will set the `clean` flag (and the `silent` flag will have its default value of false). I don't think this is an issue though since the silent flag is only used in one place (when the list of sub matrices is empty) and thus very unlikely to have ever been set (plus, in gstlearn itself `ProjMulti` is not used directly!).

Note however that the behaviour of `ProjMultiMatrix` is not changed, so I don't think it's a problem.

Overall, I would prefer if either the default was for objects to delete their components (rather than having to explicitly pass such a flag), or if the ownership (who must delete what?) was made more obvious. At the moment it is very hard to know if a pointer passed when creating an object is owned by the object or not, and if that pointer is requested to be valid for the whole life span of the object or not (sometimes it's only needed in the constructor). This is making memory management harder than it should be...

### DIFF
--- a/include/LinearOp/ProjMulti.hpp
+++ b/include/LinearOp/ProjMulti.hpp
@@ -20,7 +20,7 @@ namespace gstlrn
 class GSTLEARN_EXPORT ProjMulti: public IProj, public AStringable
 {
 public:
-  ProjMulti(const std::vector<std::vector<const IProj*>>& projs, bool silent = false);
+  ProjMulti(const std::vector<std::vector<const IProj*>>& projs, bool toClean = false, bool silent = false);
 
   /// AStringable Interface
   String toString(const AStringFormat* strfmt = nullptr) const override;
@@ -42,7 +42,7 @@ protected:
 private:
   bool _checkArg(const std::vector<std::vector<const IProj*>>& projs) const;
   void _init();
-  virtual void _clear() {};
+  virtual void _clear();
 
 protected:
   Id findFirstNoNullOnRow(Id j) const;
@@ -60,6 +60,7 @@ private:
   Id _nvariable;
   VectorInt _pointNumbers;
   VectorInt _apexNumbers;
+  bool _toClean;
   bool _silent;
   mutable VectorDouble _work;
   mutable VectorDouble _workmesh;
--- a/include/LinearOp/ProjMultiMatrix.hpp
+++ b/include/LinearOp/ProjMultiMatrix.hpp
@@ -46,9 +46,5 @@ protected:
 
 private:
   MatrixSparse _Proj;
-  void _clear() override;
-
-private:
-  bool _toClean;
 };
 } // namespace gstlrn
--- a/src/Covariances/ACov.cpp
+++ b/src/Covariances/ACov.cpp
@@ -90,6 +90,7 @@ ACov& ACov::operator=(const ACov& r)
     _pAux                  = r._pAux;
     _pw1                   = r._pw1;
     _pw2                   = r._pw2;
+    delete _tabNoStat;
     _tabNoStat             = r._tabNoStat->clone();
   }
   return *this;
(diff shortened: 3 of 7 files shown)


remi-moyen:
I had another look at this PR and I'm a bit confused, so I'm going to dump all my confusion here! 😉

According to email discussions, the philosophy is that whoever creates an object (e.g., an `IProj`) is responsible for deleting it.

This is clearly not what my PR does which would argue for rejecting it (note that there are actually 2 commits and the other one is I think fine, but it could be moved in another PR).

But this also is the opposite of what `ProjMultiMatrix` does with the `toClean` flag, so I think it would make sense to remove this flag.

However, this flag is used in the static function `ProjMultiMatrix::createFromDbAndMeshes()` which creates a bunch of `ProjMatrix` internally, and then tells the resulting `ProjMultiMatrix` that it is responsible for deleting them. This is exactly the logic of my PR, that is to have the option to make the `ProjMulti` (or `ProjMultiMatrix`) responsible for deleting its members!

Since my PR has this behaviour controlled by a flag (I've moved the `toClean` flag from `ProjMultiMatrix` up to `ProjMulti`), I feel that my PR is actually consistent with the current code of `ProjMultiMatrix`.

However we might still want to remove the `toClean` flag to make the whole code more consistent with the overall philosophy. I think this can be achieved, because `ProjMultiMatrix` does not actually uses the individual `ProjMatrix` beyond the constructor (they are all glued together in a single `SparseMatrix`). So we could remove the flag and have `ProjMultiMatrix::createFromDbAndMeshes()` delete the individual `ProjMatrix` at the end (and to avoid dangling pointers, set the `_projs` variable to an empty list).

This would require making sure that the `_projs` variable (from the base class!) is not used anywhere when using `ProjMultiMatrix`. I think this is the case, all uses (in `ProjMulti`) are in functions that are either only called from the constructor (e.g., `_checkArg()`, `_init()`...) or that are reimplemented in `ProjMultiMatrix` (`_addMesh2point()`, `_addPoint2mesh()`). The only exception is in `toString()` which could either be overridden in `ProjMultiMatrix`, or simply adapted to handle that case (there is already a cast in the derived type, which I think isn't very clean btw, so I think overriding and removing that cast would be better).

To sum up, I see three options:
1) Do nothing (reject this PR), but then we still have `toClean` in `ProjMultiMatrix` meaning it behaves differently from other pieces of code, so that's not very consistent. But it's definitely the easiest solution...
2) Reject this PR and do another one to remove `toClean` from `ProjMultiMatrix`. This means changing a bit both `ProjMultiMatrix` and `ProjMulti` to first delete the `ProjMatrix` created in `createFromDbAndMeshes()` and second make sure the `_projs` variable is never accessed when not valid.
3) Accept this PR (= move `toClean` from `ProjMultiMatrix` up to `ProjMulti`), as a way to let the user control who is responsible for deleting the `IProj`. This is slightly inconsistent with the global philosophy but it is consistent with the current code of `ProjMultiMatrix` so I don't feel it's a big problem.

Which solution sounds the best to you?

Finally, note that `ProjComposition` at the moment always deletes its members. It will need to be updated but that depends on which option we pick. For 2 (and 1, I think), it should just *not* delete its member and let that responsibility to the caller. For 3, I think a `toClean` flag (with a default of `false`) should be added to it.

fabien-ors:
@remi-moyen
Are you OK with the 2 fixes proposed in #862 and #863 ?
Can we close this PR?

remi-moyen:
The other changes should replace this PR, in a better way, so I'm closing this one.

PR #819, opened 2026-01-15
Fix memory leak in FFT

remi-moyen:
This was detected by valgrind, the internal static arrays are not deleted and thus left dangling at the end of execution.

This is a fairly small memory leak, or more importantly it's one that only happens once no matter how many times the function is called, since this is about static variables. And also, by fixing it the code might actually become slower because every successive call will always reallocate memory, whereas before if the old size was correct it wouldn't do anything.

Added to the fact that this is very old code that is (presumably) used in many places, that makes me wonder if you really want to accept this change.

On the other hand, the intent in the code above the bit that I changed is very clear that memory cleaning should happen at every call, so to me this very much look like this was a bug.

Anyway, feel free to reject the PR if you prefer.

### DIFF
--- a/src/Core/fft.cpp
+++ b/src/Core/fft.cpp
@@ -1006,6 +1006,7 @@ Id fftn(Id ndim,
       Im_Data(i) *= scaling;
     }
   }
+  fft_free(); /* free-up memory */
   return 0;
 
 Dimension_Error:


pierre-guillou:
Hi Rémi,

I took the liberty in #844 to go a little further and replace the global/`thread_local` pointers with `std::vector`s, thus making `fft_free` irrelevant.
What do you think?

remi-moyen:
Hi @pierre-guillou,

I think your change is even better than mine. I didn't want to change the code too much (or bother understanding all of its details, to be honest) so I did the minimum, but I agree that using modern containers is way better than raw pointers.

I guess you can close this PR without accepting it then?

pierre-guillou:
Thanks for your help for exhibiting the issue in the first place. I hope #844 will fix this memory leak.

Closing in favor of #844.

PR #822, opened 2026-01-16
Amelioration de Mo-Option.py

drenard77810:
This is the second pass of the improvements due to Marimo (thanks to Pierre discoveries).
It is considered as Draft until the other Mo files have been upgraded or suppressed.


fabien-ors (approved):
Thx Didier

PR #825, opened 2026-01-20
Fix negatively correlated simulations

drenard77810:
This branch was meant to correct a SEVERE bug in the multivariate simulations using Turning Bands. The correlation was not correctly handled.

In the mean times, some improvements have been achieved regarding the representation of Covariance (not symmetric variograms this time) in Python or in R.


pierre-guillou (approved):
Seems good to me

PR #826, opened 2026-01-21
Make FFT thread-safe by changing all static variables to thread_local

remi-moyen:
The FFT uses `static` variables meaning that if several threads are using it in parallel, they will overwrite each other.

Before fixing the memory leak in PR #819 this was not visible (but the issue was present!), but after that change the pointers are reset to `NULL` which sometimes causes other threads to try and access them, and crash.


pierre-guillou (approved):
Looks good to me.
Thanks Rémi!

PR #827, opened 2026-01-21
Allow using stencils to build the PrecisionOp when the model is non-stationary in sills

remi-moyen:
`ShiftOpStencils` only require stationarity in anisotropy and handles non-stationary sills, so use it when possible.


NDesassis (approved):
Nice

PR #831, opened 2026-01-22
Fix pandas >=3.0 compatibility

fabien-ors:
Fix pandas >=3.0 compatibility:
- pandas types are shorter than before: pandas.core.frame.DataFrame => pandas.Dataframe
- due to Copy-on-Write generalization, flagLocator in Db_toTL function has been discarded (see https://pandas.pydata.org/pandas-docs/stable/whatsnew/v3.0.0.)


pierre-guillou (approved):
lgtm

PR #833, opened 2026-01-22
Marimo editor width: No need for a custom CSS

pierre-guillou:
According to https://docs.marimo.io/guides/editor_features/overview/#editor-widths, marimo editor width can be set directly as a parameter to the `marimo.App` class.

This PR replaces the `custom.css` file with this parameter.

Pierre


fabien-ors (approved):
lgtm

PR #841, opened 2026-01-27
AIC is now incorporated in CovBase

drenard77810:
This branch is dedicated to suppressing the AIC (square root of the sill matrix) as a free element.
It is now incorporated as a member within the CovBase structure.
Its impact:
- in the simulation procedure (visible currently in the Simulation using Turning Bands method)
- in the automatic fitting of the sills (which is performed through the fitting of the AIC matrix... easier as it does not have to carry the positive definitness condition)

### DIFF
--- a/src/Simulation/CalcSimuTurningBands.cpp
+++ b/src/Simulation/CalcSimuTurningBands.cpp
@@ -30,8 +30,6 @@
 namespace gstlrn
 {
 
-bool _internalNewStyle = true;
-
 CalcSimuTurningBands::CalcSimuTurningBands(Id nbsimu,
                                            Id nbtuba,
                                            bool flag_check,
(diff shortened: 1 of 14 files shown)


drenard77810:
Note. Using the _aic member of Covbase in the model Fitting is finally not done. The reason is that the aic array which is handled there is local. In particular, its dimension varies along time (it is not constantly equal to nvar).
This is the reason why we renamed it into aicLocal.

src/Simulation/CalcSimuTurningBands.cpp
```diff
@@ -30,6 +29,9 @@
 
 namespace gstlrn
 {
+
+bool _internalNewStyle = true;
```
fabien-ors: Unused. To be removed?

fabien-ors (changes requested):
Just one little remark.

fabien-ors (approved):
Thx Didier

PR #842, opened 2026-01-27
GitHub Actions: Use Ubuntu for clang-check jobs

pierre-guillou:
The `check_code` workflow launches a few (8) `clang-check` jobs that ensure `gstlearn` builds on different configurations (with or without HDF5, with or without Boost `span`) without warnings.

The jobs are currently run under `macos-latest` for improved performance. Unfortunately, this slows down the whole CI since the number of parallel `macos` jobs on GitHub Actions for a given project is limited to 5.

In order to circumvent this limitation, this PR switches back to `ubuntu-latest` for these jobs. They are a few minutes of run-time slower (15 min instead of 12) but there is no limitation on their number (for now).

This should make the CI faster, especially if several workflows (from several PR) are launched concurrently.

Pierre

### DIFF
--- a/.github/workflows/check_code.yml
+++ b/.github/workflows/check_code.yml
@@ -80,7 +80,7 @@ jobs:
   # Check compiler warnings #
   # ----------------------- #
   check-warnings:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


fabien-ors (approved):
Nice idea. Thx!

PR #843, opened 2026-01-28
Correction bug de Meryem

drenard77810:
This branch has been created in order to face and solve the problem encountered by Meryem. The Kriging was producing lots of messages of 2 kinds: A zero-dimensioned array (not expected), a negative variance.

Several few improvements have been achieved here:
- the essential one is that we are now able, while using Eigen library, to check if the inversion of the Kriging matrix is feasible or not (this was the reason of the negative variances)
- a test has been added to forbid establishing the Kriging matrix as soon as the number of data is smaller than the number of drift functions
- the return code of the inversion is tested all the way through in the stack of steps used for Kriging

Therefore, in the current version, a message is produced to warn the user that, due to the choice of neighborhood, of Model (essentially the number of Drift conditions), Kriging cannot be performed. We should consider possibly suppressing the message (a similar message is simply not produced when the number of samples in the Neihborhood is zero).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,8 @@ This file contains the logs between consecutive releases
 
 Release 1.11.0
 ==============
-
+- Adding test to avoid Kriging with too few data (KrigingSimpleCase)
+- Adding a test to check that Matrix Inversion was performed correctly
 
 Release 1.10.1
 ==============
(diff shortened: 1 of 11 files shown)


fabien-ors (approved):
Thx Didier!

PR #844, opened 2026-01-29
FFT: Use std::vectors instead of void* buffers

pierre-guillou:
This PR replaces `void *` global buffers with `std::vector`s. The `void *` pointers seem to have been used for type-erasure, i.e. with either `double` or `float` data. Currently, only `double` is used.

Using `std::vector` also remove the need for freeing manually allocated memory.


fabien-ors (approved):
lgtm

PR #845, opened 2026-01-29
Correction bug calcul de covariance

drenard77810:
Quick branch to correct the calculation of covariance (instead of variogram) on a dataset containing one or several TEST values. This problem was simply lying in the calculation of the global statistics, for getting the mean and variance used in the ergodic case: the TEST values were not excluded.


fabien-ors (approved):
lgtm

NDesassis (approved):
Nice! Thx

PR #846, opened 2026-01-29
Fix MacOS architectures used for publishing R and python packages

fabien-ors:
Only two MacOs architectures are needed for compiling and publishing python and R packages (arm64 and intel [x86_64])
Publish workflows have been simplified (and accelerated) using the oldest MacOS github runners available for each of these architectures. Coverage tests still check whether all MacOS github runners can install and use the delivered packages.


pierre-guillou (approved):
lgtm

PR #847, opened 2026-01-30
Potential spde

garaytc:
\+ Add verbose for SPDE conjugate gradient
\+ Add static method to create ProjMatrix from MatrixSparse in TL
\+ Expose getMeshApicesFromCoordinates for TL

### DIFF
--- a/src/Mesh/AMesh.cpp
+++ b/src/Mesh/AMesh.cpp
@@ -488,6 +488,83 @@ VectorDouble AMesh::getApexCoordinates(Id iapex) const
   return vec;
 }
 
+VectorInt AMesh::getMeshApicesFromCoordinates(const VectorDouble& coords) const
+{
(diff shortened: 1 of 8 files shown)


src/Mesh/AMesh.cpp
```diff
@@ -789,7 +866,7 @@ void AMesh::resetProjFromDb(ProjMatrix* m,
   Ball ball(this, 10, true);
```
fabien-ors: This Ball object is no more needed here.
To be removed

fabien-ors (changes requested):
Thank you Charlie !
Just a little change requested.
++

src/Mesh/AMesh.cpp
garaytc: Done

fabien-ors (approved):
Thx Charlie!

PR #848, opened 2026-01-30
ShiftOpStencil optimization (multithreading)

remi-moyen:
Two commits:
- one minor optimisation by avoiding using a variable and multiplying by it when it is equal to 0 (doesn't actually change anything on performances)
- use OpenMP parallelisation to compute the contribution of `ShiftOpStencil`

On my test data the results are still exactly identical, regardless of whether parallelisation is used or not.

### DIFF
--- a/src/LinearOp/ShiftOpStencil.cpp
+++ b/src/LinearOp/ShiftOpStencil.cpp
@@ -13,11 +13,14 @@
 #include "Basic/Grid.hpp"
 #include "Basic/Indirection.hpp"
 #include "Basic/VectorNumT.hpp"
+#include "Basic/OptCustom.hpp"
 #include "Covariances/CovAniso.hpp"
 #include "LinearOp/AShiftOp.hpp"
 #include "LinearOp/ShiftOpMatrix.hpp"
 #include "Mesh/MeshETurbo.hpp"
 #include "geoslib_define.h"
+
+#include <omp.h>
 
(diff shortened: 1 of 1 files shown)


pierre-guillou:
Hi Rémi,

I rebased on top of `origin/dev` and added two small modifications:
- I moved the `total` variable inside the parallel loops to avoid `private(total)`
- I replaced the outer `parallel` block with the `firstprivate` clause.

@remi-moyen: Do my changes still work on your test data? If not, I will remove them.
Pierre

remi-moyen (commented):
@pierre-guillou your changes seem to work.

I have no issue with the first one (moving `total` inside the loop), it's possibly a tiny bit less efficient but unlikely to matter. I was a bit less confident on the second one since `private()` caused a crash for me but it seems `firstprivate()` does work correctly. I'm not sure why (each thread working from an uninitialised object should be the same as working from a copy of the default uninitialised one!), but all that matters is that it works.

Also github doesn't let me approve my own PR so I can't do more than this comment.

pierre-guillou:
Thanks for checking!

The main difference between `private` and `firstprivate` is, to my knowledge, that `private` will default-initialize the variables. `firstprivate` will make a copy. With `private`, you might have needed to resize the vectors, which should be unnecessary with `firstprivate`.

pierre-guillou (approved):

PR #849, opened 2026-02-04
Fix bug in MatrixSparse using a 32 bits index

remi-moyen:
The code uses `Eigen::SparseMatrix<double>` which implicitly means the `StorageIndex` is an `int` (see `Eigen/src/SparseCore/SparseUtil.h`, line 52).

The fix is to specify the `StorageIndex` everywhere (this was already done e.g., in `NF_Triplet.hpp` when defining the `Eigen::Triplet`), for convenience there is a new `typedef EigenSparseMatrix` used in several places.


pierre-guillou (approved):
Thanks Rémi!

PR #851, opened 2026-02-05
Refactoring simulations

drenard77810:
This branch is a major step in the reformulation of Gaussian simulations.
The work is not finished yet. But a major step is performed as the Spectral simulations are now presented at an equivalent stage as Turning Bands ones.
Moreover both simulations have bee embedded in a Calculator framework.

The final step will be:
- to polish the new classes dedicated to Spectral simulations (essentially by mimicking the other classes embedded in calculators ... such as Kriging for example)
- this will allow proposing conditional simulations, whatever the engine (TB or Spectral)

Note one difference nevertheless: TB is restricted to Model with list of CovAniso covariances, whereas Spectral is more open as it can cope with ModelGeneric (no restriction).

Possibly, we could even propose a new API for perforing conditional or non-conditional simulation, where the engine choice (TB or Spectral) will be defined as an argument (this need should be discussed first within the team).


doc/demo/r/Tuto_SimuSpectral.Rmd
```diff
@@ -58,25 +58,24 @@ dx   = c(1.)
 grid = DbGrid_create(nx = nx, dx = dx)
 
 # specification of the model using Model
-mod  = Model_createFromParam(type = ECov_fromKey(type_cov), 
+mod  = Model_createFromParam(type = ECov_fromKey(type_cov), range=1,
```
fabien-ors, line 61: One out of "range" or "ranges" is unused.

fabien-ors (approved):
I didn't reviewed all files.

PR #853, opened 2026-02-06
Global refactor simulations

drenard77810:
This branch is the second phase for restructuring the simulations. 
For the time being, the first stage is just a cleaning one, where the number of shadowed variables has been almost reduced to none (this avoids late errors due to possible warnings on Windows which are mutated to errors and therefore interrupt any PR).
But we are far from completion. So this branch is considered as a Draft for the time being.


fabien-ors (approved):
Thx Didier

PR #857, opened 2026-02-09
ASerializable: Set locale to support accents in Windows paths

pierre-guillou:
On Windows, gstlearn currently cannot read or write files in a filesystem hierarchy using non-ASCII characters. This was detected by running `Mo-Option.py` inside a directory named with accented characters.

This PR fixes this issue by setting the locale in the `ASerializable` constructor, a class used to read & write files.

Pierre


fabien-ors (approved):
lgtm

PR #859, opened 2026-02-10
Correction

### LINKED ISSUE #858: Infer space in Model::createFromParam
`model = gl.Model.createFromParam(gl.ECov.MATERN,ranges = [3.,4.,5.])`

gives a message 

_Incompatibility between:
Space Dimension = 2
Dimension of argument 'ranges' = 3_

and doesn't create a Model.

It is required to do 
`sp = gl.SpaceRN.create(3)`
`model = gl.Model.createFromParam(gl.ECov.MATERN,ranges = [3.,4.,5.], space = sp)`

Wouldn't it possible to infer the dimension and to create the space internally when the user seems to require a specific dimension (when providing a vector of ranges)?

drenard77810:
When using a default space of Dimension 2. 
If you attempt to create a new model using createFromParam() function, you can specify some anisotropy by giving a vector of ranges. The dimension of this vector should give an information on the space dimension.
If this range-dimension is not equal to the default space dimension, a message was issued.

Instead, after this correction, the newly created model will have the space dimension provided by the length of the range vector. 
However, an error is still issued if the size of the range vector is not in accordance with the space argument that can also be provided as an argument to the same function.

Fix #858

### DIFF
--- a/src/Model/Model.cpp
+++ b/src/Model/Model.cpp
@@ -146,11 +146,14 @@ Model* Model::createFromParam(const ECov& type,
   Id nvar = 1;
   if (!sills.empty()) nvar = sills.getNRows();
 
-  Id ndimRanges = (ranges.empty()) ? 0 : static_cast<Id>(ranges.size());
+  Id ndimRanges = 1; // By default, the argument 'range' gives the space dimension.
+  // If ranges is provided, its length provides an infirmation on the space dimension
+  if (!ranges.empty()) ndimRanges = static_cast<Id>(ranges.size());
 
   CovContext ctxt;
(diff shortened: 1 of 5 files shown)


src/Model/Model.cpp
fabien-ors, line 1: Could it be possible to add a doxygen documentation to the function ?

fabien-ors (changes requested):
Thx Didier !

drenard77810:
I think that the requested modifications have been done. I probably need to have Fabien validate the changes.

fabien-ors (approved):
Thx Didier

PR #861, opened 2026-02-13
Cosmetic in python course (vario)

fabien-ors:
Fix a comment in 04_Variography.ipynb (plot.varmod is an R function)


pierre-guillou (approved):
lgtm

PR #862, opened 2026-02-13
ACov: Wrap _tabNoStat in std::unique_ptr

pierre-guillou:
This PR wraps the `_tabNoStat` pointer in `ACov` in `std::unique_ptr`. This way, the memory is automatically freed when the `ACov` instance is destroyed.

Also, `ACov` methods returning pointers to this member were modified to return references. This makes memory handling more explicit.

This PR should also solve the first memory issue in #818.

Pierre


fabien-ors (approved):
lgtm

PR #863, opened 2026-02-13
ProjMulti: Use std::optional<reference> instead of pointer

pierre-guillou:
This PR reworks the `_projs` member of the `ProjMulti` class to make sure memory is externally managed. To do so, naked pointers have been replaced with references wrapped in `std::optional` (because the naked pointers were allowed to be `nullptr`).

Work has been done to adapt `ProjMultiMatrix` to this change. In particular, since

- the `ProjMultiMatrix::createFromDbAndMeshes` method did transfer the ownership to the `IProj` variables to the `ProjMulti` instance in the `_projs` variable
- while the `ProjMultiMatrix`  constructor kept the ownership of `_projs` outside the instance.

To account for these two behaviors, a new member variable in `ProjMultiMatrix` named `_projsStore` owns the memory of the `IProj`s when called from `createFromDbAndMeshes`. It is empty when the main constructor is used.

This should solve the second memory leak of #818 (the typo fix contained in this PR has also been transferred here).

Pierre

### DIFF
--- a/include/LinearOp/ProjComposition.hpp
+++ b/include/LinearOp/ProjComposition.hpp
@@ -13,7 +13,7 @@
 #include "LinearOp/IProj.hpp"
 #include "gstlearn_export.hpp"
 
-#include <memory>
+#include <functional>
 
 namespace gstlrn
 {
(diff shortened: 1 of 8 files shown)


fabien-ors:
@remi-moyen Should ProjComposition class adopt the same logic ?
(see #818 discussion)

pierre-guillou:
You're right, the way `ProjComposition` is currently coded, it will take ownership of its input `std::vector<const IProj*> projs`. I fear this does not play well with the Python wrapper. I don't see it used in any Python script though.

remi-moyen (approved):
Sounds good to me.

I'm probably the only one using `ProjComposition` which is why it's not in any Python script.

PR #872, opened 2026-02-18
Fix instructions for installing gstlearn under R

fabien-ors:
Replace install.package instruction by install.packages in R README.


pierre-guillou (approved):

NDesassis (approved):
Thx!

PR #873, opened 2026-02-19
SPDE on 1D, inform anistropies on meshes/vertices, and bug fix

garaytc:
+ Add a workaround for SPDE to work in 1D
+ Add a flag to ACov/CovAniso to inform local anistropies directly on meshes/vertices
+ Expose TildeC vector

--
+ Fix in getMeshAndInPlaceWeightsFromCoordinates(), for Ball Tree search, when the number of meshes is lower than the number of defined neighbors

### DIFF
--- a/include/Covariances/ACov.hpp
+++ b/include/Covariances/ACov.hpp
@@ -458,6 +458,9 @@ public:
     return TEST;
   }
   void makeStationary();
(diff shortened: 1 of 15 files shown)


NDesassis (changes requested):
ACov shouldn't be aware about anisotropy. It is something only related to CorAniso. Since I know this features are for SPDE (which only deals with CorAniso) , it is possible to limit the reference to Aniso  (member + methods) to CorAniso.

garaytc:
Done @NDesassis

NDesassis (approved):

PR #874, opened 2026-02-20
Add explanation when no internet is detected

fabien-ors:
Executing loadData function (either in R or python) could fail if:
- the requested data file is not located in sub-directories (add verbose=True) - AND -
- an Internet connection is not detected (all tutorial data files are located on our website gstlearn.org)

To test if an Internet connection is available, we use:
- 'requests' package in Python
- 'lares' package in R

When loadData fails, the "No Internet" error message is now completed by a sentence about those required packages.

### DIFF
--- a/r/modules/document.R
+++ b/r/modules/document.R
@@ -114,7 +114,7 @@ locateFile <- function (filename, where='references', directory=NULL, verbose=FA
   # Test on the web
   if (!internetAvailable())
   {
-    print(paste("Error: Cannot access to", filename, "(no Internet)!"))
+    print(paste("Error: Cannot access to", filename, "(no Internet - 'lares' package is required)!"))
     return(NULL)
   }
   


python/modules/document.py
```diff
@@ -221,7 +221,11 @@ def locateFile(
 
     # Test on the web
     if not internetAvailable():
-        print("Error: Cannot access to", filename, "(no Internet)!")
+        print(
+            "Error: Cannot access to",
+            filename,
+            "(no Internet - 'requests' package is required)!",
```
pierre-guillou: If `requests` is missing, importing `document.py` fails with `ImportError` that tells users to install `gstlearn[doc]`.
This function should not be callable without `requests`.

fabien-ors: Ok. This is not the case in R. I rolled back the modification in python

pierre-guillou:
The second commit should be squashed into the first one.

fabien-ors:
squash done

pierre-guillou (approved):

PR #876, opened 2026-02-20
Testing new operator overload for VectorDouble

drenard77810:
THis branch has been created very temporarily in order to check the new format that will be used for some classes.
We have tested it on the method sum() currently in the class VectorNumT.
It has been moved as free fucntions in VectorHelper.
It has been renamed into add()
The number of different prototypes has been reduced drastically
The overload of operators has been tested.
The convincing result is integrated into the cpp test called: test_VectorHelper.


src/Basic/VectorHelper.cpp
```diff
@@ -1911,4 +1911,37 @@ void VectorHelper::capInPlaceVVD(VectorVectorDouble& vec, double vmin, double vm
...
+void VectorHelper::add(VectorDouble& vecout, const VectorDouble& v1, const VectorDouble& v2)
+{
+  auto n1 = static_cast<Id>(v1.size());
+  auto n2 = static_cast<Id>(v2.size());
+  if (n1 != n2)
+  {
+    messerr("Impossible to add vectors of different dimensions\n");
+    return;
+  }
+  if (&vecout != &v1 && &vecout != &v2)
```
fabien-ors: why don't testing size of vecout only?

`if (vecout.size() != v1.size())`

include/Basic/VectorHelper.hpp
```diff
@@ -239,11 +239,30 @@ class GSTLEARN_EXPORT VectorHelper
   static VectorDouble compress(const VectorDouble& vecin, const VectorInt& vindex);
   static void truncateDecimalsInPlace(VectorDouble& vec, Id ndec);
   static void truncateDigitsInPlace(VectorDouble& vec, Id ndec);
+
+  static VectorDouble add(const VectorDouble& v1, const VectorDouble& v2);
```
fabien-ors: Why don't using a template function as before?

fabien-ors (changes requested):

drenard77810:
After discussion with Nicolas, it appears that using a template:
- can always be done afterwards if necessary
- currently, using template is only used in order to reduce the number of codes when using the same function (say add) for different types.
Looking honestly, it appears that the only OTHER types that could be of interest are: VectorInt and VectorVectorDouble.

But then, the question could be:
- do we really have to make functions more difficult to grasp in order to offer the possibility to apply the methods to VectorInt (this type is only used to manipulate indices ... so what would be the meaning of add() methhod applied to this type).
- Similar question for VectoVectorDouble.
As a (may be temporary) decision, we imagined to concentrate to translate the maximum number of functions (from VectorNumt to VectorHelper) first, to update the code consequently. The compiler should produce the places where the functions were actually used for VectorInt and VectorVectorDouble.
If too many, we would then think to integrate the template back. Otherwise, we can still imagine to develop a addVVD() specifically if needed.

fabien-ors:
A template for adding VectorNumT element-wise can be applied for any vectors of numerical values (int, double, float, etc...). Vector of vectors are out of the scope in that case. But yes, the template function could be added later of course if really needed.

NDesassis (approved):

src/Basic/VectorHelper.cpp
```diff
@@ -724,10 +778,7 @@ void VectorHelper::addSquareInPlace(VectorDouble& dest, const VectorDouble& src)
   }
 }
 
-void VectorHelper::addInPlace(const double* veca,
-                              const double* vecb,
-                              double* res,
-                              Id size)
+void VectorHelper::addInPlace(const double* veca, const double* vecb, double* res, Id size)
```
pierre-guillou, line 766: Use `span` / `consvect` instead of raw pointers + signed size

doc/courses/python/08_SimuPGS.ipynb
pierre-guillou, line 1: Useless changes

src/Estimation/CalcImage.cpp
```diff
@@ -250,12 +251,10 @@ DbGrid* CalcImage::_buildMarpat(const NeighImage* neigh,
   VectorInt center = dbgrid->getCenterIndices();
 
   // Loop on the valid weights
-  VectorInt local(ndim);
   for (Id ineigh = 0; ineigh < nbneigh; ineigh++)
   {
-    local = ranks[ineigh];
-    local.add(center);
-    Id iadd = dbgrid->indiceToRank(local);
+    auto local = ranks[ineigh] + center;
```
pierre-guillou: Doesn't this add an allocation per loop iteration?

src/LinearOp/ShiftOpStencil.cpp
```diff
@@ -133,8 +134,7 @@ Id ShiftOpStencil::_addToDest(const constvect inv, vect outv) const
         grid.rankToIndice(rank, center);
         for (Id iw = 0; iw < nw; iw++)
         {
-          local = center;
-          local.add(_relativeShifts[iw]);
+          local = center + _relativeShifts[iw];
```
pierre-guillou: Is the `local` vector re-allocated at each loop iteration?

src/LinearOp/ShiftOpStencil.cpp
```diff
@@ -273,8 +273,7 @@ Id ShiftOpStencil::_buildInternal(const MeshETurbo* mesh,
 
   for (Id iw = 0; iw < nw; iw++)
   {
-    VectorInt local = center;
-    local.add(_relativeShifts[iw]);
+    VectorInt local     = center + _relativeShifts[iw];
```
pierre-guillou: Surely we can factor `local` out of the for loop

include/Basic/VectorHelper.hpp
```diff
@@ -239,11 +242,572 @@ class GSTLEARN_EXPORT VectorHelper
   static VectorDouble compress(const VectorDouble& vecin, const VectorInt& vindex);
   static void truncateDecimalsInPlace(VectorDouble& vec, Id ndec);
   static void truncateDigitsInPlace(VectorDouble& vec, Id ndec);
+
```
pierre-guillou, line 245: Should Target-Language specific code be moved into `swig_inc.i`?

include/Basic/VectorHelper.hpp
```diff
@@ -239,11 +242,572 @@ class GSTLEARN_EXPORT VectorHelper
...
+  VectorHelper::add(result, v1, v2);
+  return result;
+}
+template<typename T, typename U>
+inline VectorNumT<T> operator+(const VectorNumT<T>& v1, U v2)
+{
+  static_assert(std::is_same_v<T, U> ||
+                  (std::is_same_v<T, Id> && std::is_integral_v<U>) ||
+                  (std::is_same_v<T, double> && std::is_arithmetic_v<U>),
+                "Invalid type for addition operator between VectorNumT and scalar");
```
pierre-guillou: Why is this check outside VectorHelper::add?

include/Model/Model.hpp
```diff
@@ -161,6 +161,7 @@ class GSTLEARN_EXPORT Model: public AStringable, public ModelCovList
   FORWARD_METHOD_NON_CONST(_castInCovAnisoList, setActiveFactor)
   FORWARD_METHOD_NON_CONST(_castInCovAnisoList, getCovAniso)
   FORWARD_METHOD_NON_CONST(_castInCovAnisoList, setRangeIsotropic)
+  FORWARD_METHOD_NON_CONST(_castInCovAnisoList, setType)
```
pierre-guillou, line 164: How is this change related to the current PR?

src/Basic/VectorHelper.cpp
```diff
@@ -490,6 +490,60 @@ VectorDouble VectorHelper::sequenceVD(double valFrom,
   return vec;
 }
 
+VectorBool VectorHelper::simulateBoolean(Id n, double probaTrue)
```
pierre-guillou, line 493: This method allocates

src/Basic/VectorHelper.cpp
```diff
@@ -490,6 +490,60 @@ VectorDouble VectorHelper::sequenceVD(double valFrom,
...
+{
+  VectorBool vec(n);
+  auto it(vec.begin());
+  while (it < vec.end())
```
pierre-guillou: Why a `while` loop instead of a range-based for loop?

src/Basic/VectorHelper.cpp
```diff
@@ -490,6 +490,60 @@ VectorDouble VectorHelper::sequenceVD(double valFrom,
...
+    messerr("The sum of probabilities should be positive. Nothing is done.");
+    return VectorInt();
+  }
+  auto nproba = static_cast<Id>(probas.size());
+  VectorDouble normprobas(nproba);
+  for (Id i = 0; i < nproba; i++)
+    normprobas[i] = probas[i] / total;
+
+  auto it(vec.begin());
+  while (it < vec.end())
```
pierre-guillou: Could be changed into a range-based for loop

include/Basic/VectorHelper.hpp
```diff
@@ -239,11 +242,572 @@ class GSTLEARN_EXPORT VectorHelper
...
+    for (Id i = 0; i < n1; i++)
+      vecout[i] = (isNA(v1[i]) || isNA(v2[i])) ? getNA<T>() : v1[i] + v2[i];
+  }
+  template<typename T>
+  static void increment(VectorNumT<T>& vecout, const VectorNumT<T>& v1)
+  {
+    VectorHelper::add(vecout, vecout, v1);
+  }
+  template<typename T>
+  static void add(VectorNumT<T>& vecout, const VectorNumT<T>& v1, T v2)
```
pierre-guillou: Why not also templating the scalar variable too?

pierre-guillou (changes requested):

src/Estimation/CalcImage.cpp
drenard77810: Done: use the inplace method

src/LinearOp/ShiftOpStencil.cpp
drenard77810: done (use the inplace method)

src/LinearOp/ShiftOpStencil.cpp
drenard77810: Done: use the in place method

include/Model/Model.hpp
drenard77810: I need to have this line modified because I needed this in my course (done simultaneously).

src/Basic/VectorHelper.cpp
drenard77810: Yes. Like all the other simulateXX methods of VectorHelper.

src/Basic/VectorHelper.cpp
drenard77810: I thought it was better programming. I return to the range based traditional loop.

src/Basic/VectorHelper.cpp
drenard77810: Done

include/Basic/VectorHelper.hpp
drenard77810: I did not understand the remark. 
The scalar IS templated already...
You meant: it could be templated separately???

src/Basic/VectorHelper.cpp
```diff
@@ -490,6 +490,54 @@ VectorDouble VectorHelper::sequenceVD(double valFrom,
   return vec;
 }
 
+VectorBool VectorHelper::simulateBoolean(Id n, double probaTrue)
+{
+  VectorBool vec(n);
+  for (Id i = 0; i < n; i++)
```
pierre-guillou: This is a classical for loop, not a range-based for loop, which looks like
```cpp
for(auto &el: vec) {
  el = law_uniform(0., 1.) < probaTrue;
}
```

drenard77810: OK

src/Estimation/CalcImage.cpp
drenard77810: done VH::add()

include/Basic/VectorHelper.hpp
drenard77810: Done

pierre-guillou (approved):

fabien-ors (approved):

PR #878, opened 2026-02-23
Target Python Stable ABI to produce wheels compatible with more Python versions

pierre-guillou:
This PR uses the Python Limited API / Stable ABI (https://docs.python.org/3/c-api/stable.html#stable-application-binary-interface) to generate Python packages compatible with several Python versions at once.

Only one CPython function explicitly called in SWIG typemaps has been changed to its corresponding function in the Python Limited API: `PyUnicode_AsUTF8` became `PyUnicode_AsUTF8AndSize`, which is compatible with Python 3.10 and onwards.

The `publish_python` workflows have been updated to produce only one wheel per platform. This wheel should be compatible with Python 3.10 and more recent versions. Wheel tags have been updated. Test jobs still use several Python versions since they are a lot faster.

~~The same has been done for R packages. Since building them on older R versions seem to work on latest R, the number of building jobs has been reduced to only one R, the `4.2.3`.~~

Corresponding publish workflow: https://github.com/pierre-guillou/gstlearn/actions/runs/22306793706

In addition,

- `actions/upload-artifact` has been upgraded from `v4` to `v6`
- `actions/download-artifact` has been upgraded from `v4` to `v7`
- building the Matrix R dependency is not needed anymore (maybe related to https://github.com/gstlearn/gstlearn/commit/35f18576d05b19a242e196fc1ad06e8cf4fbcfe7)
- `uv` is used in the `publish_python_windows` job
- the `x86` Windows platform for Python has been removed.

Overall, this should reduce a lot the time spend in the `publish` workflow, since we are now under the 5 jobs limitation for the macOS jobs (at least for the `build` jobs).

Pierre

### DIFF
--- a/.github/actions/compress_and_upload_nonreg_artifacts/action.yml
+++ b/.github/actions/compress_and_upload_nonreg_artifacts/action.yml
@@ -23,13 +23,13 @@ runs:
         mv ${{ inputs.os-name }}-neutral.tar.gz $GITHUB_WORKSPACE/
 
     - name: Publish output logs as artefact
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v6
       with:
         name: ${{ inputs.os-name }}-nonreg-logs
         path: ${{ inputs.build-dir }}/tests/${{ inputs.os-name }}-logs.tar.gz
 
     - name: Publish output neutral files as artefact
-      uses: actions/upload-artifact@v4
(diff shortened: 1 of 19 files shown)


fabien-ors (approved):
Thanks to my own tests :
https://github.com/fabien-ors/swigex0/actions/runs/22967740969
I approve!
Thx Pierre!

PR #882, opened 2026-02-27
Use num_threads() rather than omp_set_num_threads()

### LINKED ISSUE #881: Do not use omp_set_num_threads() in gstlearn
There are a few places in gstlearn that use OpenMP parallelization, and as far as I can see they all use `omp_set_num_threads()` to set the number of threads to use (with the value taken from `OptCustom::query("ompthreads", 1)`).

However `omp_set_num_threads()` affects the whole process, not just the function where it is called, meaning that if later another part of the code uses OpenMP, it will be affected by the value set by gstlearn. This does not matter as long as all OpenMP calls use `omp_set_num_threads()` but this not the only way to specify a number of threads to use.

I think it is far better to use `num_threads()` in the `#parallel` clause as this has the same result (the number of threads is limited within the parallel region), but without "leaking" this to other code using gstlearn.

Searching for `omp_set_num_threads()` currently returns 4 matches:
- ShiftOpStencil.cpp
- CalcKrigingSimpleCase.cpp
- MatrixSparse.cpp
- MultiGridSPDE.cpp

remi-moyen:


pierre-guillou (approved):
Thanks Rémi!
Indeed, `omp_set_num_threads` has non local effects, contrary to `num_threads`.

PR #886, opened 2026-03-17
Fix publish workflows

fabien-ors:
This new PR also sticks to scikit-sparse==0.4.16 in publish workflows


pierre-guillou (approved):
Context: #884

PR #887, opened 2026-03-18
Correction Covariance NC

drenard77810:
Despite its name, the calculation of the Non centered covariance was ... centering the variables to their respective mean, calculated either locally or globally (depending on the Ergodic flag).
The correction was simply to forbid the centering.


NDesassis (approved):
Thx Didier!

PR #897, opened 2026-03-30
Correcting the instability

drenard77810:
This bug was linked to an instability generated by the function Vario::getNext().
This function was not properly answering in the case of call for an Asymetric covariance. And this happened ONLY in the case of fitting using Covariance (as for PGS for example).
The result was unpredictable as it returned ITEST which was used next as an address.

In the mean time, the function test_a_template has been added to the list of IGNORED files to enable manipulating un-cleaned versions of this file.

Ultimately (and with no link) , the MD file Derivative.tex was cleaned out of latex_syntax_checker problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,4 +60,5 @@ test_*.NF
 test_*.out
 test_*-Data
 test_*-Grid
+test_a_template.cpp
 .ruff_cache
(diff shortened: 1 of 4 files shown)


NDesassis (approved):
Thx Didier!

PR #899, opened 2026-04-02
Update clang-format configuration

pierre-guillou:
This PR updates the `.clang-format` file with the following changes:
- `BinPackArguments` set to `true`: function calls are allowed to have several arguments on the same line
- `BinPackParameters` set to `AlwaysOnePerLine` instead of `false` (this should have the same behavior): in function declarations & definitions, keep one parameter per line
- `BreakAfterOpenBracketFunction` set to `true`: in function declarations, definitions & calls, force a line break if parameters/arguments are too long. This last option is clang-format >= 22.

`Style.hpp` has been updated accordingly.

Pierre

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,4 @@
-# This requires a version of clang-format above 21
+# This requires a version of clang-format above 22
 AlignAfterOpenBracket: true
 AlignArrayOfStructures: None
 AlignConsecutiveAssignments: None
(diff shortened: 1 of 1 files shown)


pierre-guillou:
Update: `BinPackParameters` set to `OnePerLine`, which allows some binpacking when < ColumnLimit.
No more change in `Style.hpp`.

fabien-ors (approved):
Thx Pierre !
Ok for me

PR #904, opened 2026-04-09
GitHub Actions: Fix nonreg-macos brew install error

pierre-guillou:
Ignore `brew install` error when install `gdal`.


fabien-ors (approved):
Nice point Pierre :)

PR #906, opened 2026-04-10
Enable pre-commit in GitHub Actions

pierre-guillou:
This PR enables `pre-commit` in the `check_code` workflow. A few source files were not fully formatted, the PR fixes that too.

Pierre


fabien-ors (approved):
Thx Pierre.
I forgot enabling my local pre-commit before submitting the "Reformat" PR.
It is now installed :)

PR #908, opened 2026-04-13
Replace assert in clone() with static_assert

pierre-guillou:
The assert is a runtime check that is removed in Release mode. On the contrary, static_assert is checked at compile-time, independently of the build type.

The assert has been triggered in `InvNuggetOp::cloneInvNuggetMatrix()` because `typeid(Derived) != typeid(Base)`. The static_assert fixes this issue as well.


pierre-guillou:
The `assert` was triggered by `test_buildInvnoise.py` when building gstlearn in Debug type.

fabien-ors (approved):
Thx Pierre

PR #911, opened 2026-04-14
MatrixSparse: Fix Eigen assert in Debug mode

pierre-guillou:
The `MatrixSparse::_prodnormGeneral` method is using a matrix-vector-matrix multiplication that hits an assert in Eigen when gstlearn is built in Debug mode.

The assert message is `lhs.cols() == rhs.rows() && "invalid matrix product" && "if you wanted a coeff-wise or a dot product use the respective explicit functions"`. See `test_Vecchia` for instance.

This PR fixes the issue by converting the inner vector to a diagonal matrix, which already occurs in `MatrixSparse::prodNormMatVecInPlace`.

Pierre


fabien-ors (approved):
Thx Pierre

PR #913, opened 2026-04-15
Add missing GSTLEARN_EXPORT

pierre-guillou:
Has caused an `undefined symbol` error at run-time when importing the Python package.

Pierre


fabien-ors (approved):
lgtm

PR #914, opened 2026-04-17
Virtual fault mesh + SPDE Gibbs

garaytc:
- Add a MeshEFaulted that allows to define virtual meshes, in order to handle faults with SPDE
- Add Gibbs SPDE to handle inequality constraints
- Fix law_gaussian_between_bounds if one bound is undefined and the other is higher/lower than the predefined "large" value


NDesassis (approved):
Thx Charlie!

PR #915, opened 2026-04-17
AMatrix: Avoid temp variables in linearCombinationInPlace

pierre-guillou:
This PR rewrites the `AMatrix::linearCombinationInPlace` method to remove the temporary matrices `cumul` and `temp`.

This avoids temporary allocations and deallocations.

There's a small change in the functions semantics: I rely here on the fact that `val3 != 0.` implies `val2 != 0.` and `val1 != 0.`. If not the case, more `if`s should be added.

Pierre

### DIFF
--- a/include/Matrix/AMatrix.hpp
+++ b/include/Matrix/AMatrix.hpp
@@ -468,29 +468,43 @@ namespace gstlrn
         res.resize(other1.getNRows(), other1.getNCols());
       }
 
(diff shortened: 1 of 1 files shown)


pierre-guillou:
Additional `if`s added to support more cases of `val` being `0.0`.

fabien-ors (approved):
lgtm

PR #921, opened 2026-05-05
fix case with infinite random simulations

NDesassis:
Patch to avoid case with nan in Gneiting simulation (patch made by X. Freulon).


fabien-ors (approved):
Thx Xavier and Nicolas!

PR #922, opened 2026-05-09
perf: fix broken pair-scan early-exit and add Ball tree range query to _calculateGeneralByPair

collord:
## Summary

- **Bug fix:** `getDistance1D(iech, jech)` was inverted in 6 pair-scan break conditions across `Vario.cpp`, `VarioParam.cpp`, and `variopgs.cpp`. The function computes `X[iech] - X[jech]`, which is always ≤ 0 for ascending-sorted pairs (jech has larger X), so the break never fired and all O(n²/2) pairs were evaluated regardless of `maxdist`.

- **Performance:** Replaced the 1D sorted-array early-exit in `Vario::_calculateGeneralByPair` (non-date path) with a Ball tree range query (`Ball::queryRadiusInPlace` / `t_btree::query_radius_depth_first`). This finds all candidate pairs within a true Euclidean sphere of radius `maxdist` in O(k log n) per query, rather than pruning only along the X axis.

- The date-aware path retains the sorted fallback with the corrected break condition.

## Performance

On a 9945-point 3D dataset with 3 directional variograms:

| State | Wall time |
|---|---|
| Before (broken break) | ~30.7 s |
| After break fix only | ~13.2 s |
| After break fix + Ball tree | ~5.7 s |

Total: **5.3× speedup**.

## Changes

- `src/Variogram/VarioParam.cpp`, `src/Core/variopgs.cpp`: single-line `getDistance1D` argument swap each
- `src/Variogram/Vario.cpp`: three `getDistance1D` swaps in other functions; `_calculateGeneralByPair` main loop replaced with Ball tree range query
- `include/Tree/ball_algorithm.h`, `src/Tree/ball_algorithm.cpp`: add `t_btree::query_radius_depth_first`
- `include/Tree/Ball.hpp`, `src/Tree/Ball.cpp`: add `Ball::queryRadiusInPlace`

## Test plan

- [ ] Verify existing variogram tests pass
- [ ] Confirm results match the original O(n²) path on a known dataset
- [ ] Check that the date-aware fallback path still works correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/Tree/Ball.hpp
+++ b/include/Tree/Ball.hpp
@@ -55,6 +55,7 @@ namespace gstlrn
     KNN queryOneAsVDFromSP(const SpacePoint& Pt, Id n_neighbors = 1);
     virtual VectorInt getIndices(const SpacePoint& Pt, Id n_neighbors = 1);
     Id queryClosest(const VectorDouble& test);
+    void queryRadiusInPlace(const VectorDouble& test, double radius, VectorInt& indices) const;
     virtual Id queryOneInPlace(
       const VectorDouble& test,
       Id n_neighbors,
--- a/include/Tree/Ball.hpp
+++ b/include/Tree/Ball.hpp
@@ -55,7 +55,10 @@ namespace gstlrn
     KNN queryOneAsVDFromSP(const SpacePoint& Pt, Id n_neighbors = 1);
     virtual VectorInt getIndices(const SpacePoint& Pt, Id n_neighbors = 1);
     Id queryClosest(const VectorDouble& test);
-    void queryRadiusInPlace(const VectorDouble& test, double radius, VectorInt& indices) const;
+    void queryRadiusInPlace(
+      const VectorDouble& test,
+      double radius,
+      VectorInt& indices) const;
     virtual Id queryOneInPlace(
       const VectorDouble& test,
       Id n_neighbors,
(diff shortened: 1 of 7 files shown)


fabien-ors:
Thank you very much for your contribution.

Regarding check-formatting workflow, I suggest you to install pre-commit python package in order to automatically launch the formatting job before each of your commit. To do so, from a terminal in the root folder of the repository:

```
uv pip install pre-commit
pre-commit install
```

We look forward the "Ready for review" status of your PR.
Thank you.

fabien-ors:
Thank you for the formatting.
Now we see that some non-regression tests are failing.
Do you want to try to fix them by yourself ?

collord:
> Thank you for the formatting. Now we see that some non-regression tests are failing. Do you wan to try to fix them by yourself ?

After the modified PR today I believe there are no new regressions in the tests. (If I would have known we would need to be patching ball_algorithm.cpp, I would have just submitted a PR fixing only the getDistance1D(iech, jech) issue.)

Thank you

fabien-ors:
Dear

Thank you very much for your feedback.

There is still a test that fails (SEGFAULT) under MacOS only :
https://github.com/gstlearn/gstlearn/actions/runs/25814420872/job/75860453366?pr=922
We are going to investigate.

All other failures are related to reference logs that must be updated due to your bug fix.
We are going to update the reference logs by ourself.

We keep in touch.
Thank you!

PR #928, opened 2026-06-01
Revisiting rule to add facies names and colors

drenard77810:
This branch is dedicated to the improvement of the Rule contents: now, it also stores the Name, the Color and the Assigned Value for each Facies.
This enables, in particular, to define the colors used to represent a categorical variable (obtained using a Rule-based method) without having to carry a color map and to perform the facies-color assignment by hand.
Moreover the legend becomes more legible as the correct names are reported together with the dedicated colors.

A test has been created to check the main features of the updated Rule class.
Two additional Tuto have been developed to test the new features in R and Python.


src/LithoRule/Rule.cpp
```diff
@@ -1153,6 +1268,10 @@ namespace gstlrn
       setMainNodeFromNodNames(nodes);
     }
 
+    ret = ret && SerializeHDF5::readVec(*ruleG, "FaciesNames", _facnames);
```
fabien-ors: Would it be possible to accept reading old format?
(i.e. don't fail if there is no facies colors, names and values when reading the file)

fabien-ors (changes requested):
Small change to handle retro-compatibility

src/LithoRule/Rule.cpp
drenard77810: Done

src/Core/thresh.cpp
pierre-guillou, line 1: Removing this function feels a bit weird to me, in particular when its body has been inlined in its callers.
I would suggest to wraw the `rule` pointer in the `local_pgs` variable in `variopgs.cpp` in the `testpgs.cpp` into `std::unique_ptr` to avoid manual deallocation when possible. Otherwise use `std::unique_ptr::reset()`.

fabien-ors (approved):

PR #929, opened 2026-06-02
Fix compilation with VS 2017 / boost::span

remi-moyen:
* boost::span doesn't have constructors from iterators, so use a constructor from pointer instead.
* std::remove_cvref is C++20 only, for C++17 we can use std::decay. This isn't strictly equivalent however for our use-case this is the same.
* For some reason VS refuses to use the move assignment of MultiGridSolver (in SPDEOp::SPDEOp) when it is marked noexcept (and it tries to use the copy assignment which is deleted so we get an error, which is good here!). I don't think that noexcept is very important in this declaration so I've commented it out.
* There is a crash in IPrecisionOp default functions, see comment in code. I am uneasy with this crash, and the fix, because I don't understand either of them, and I can't find a reference to any VS2017 bug that would explain it. I also wasn't able to extract a simplified minimal example that crashes. This means that issue could happen anywhere else as I can't give a rule of when it does...

### DIFF
--- a/include/Basic/ICloneable.hpp
+++ b/include/Basic/ICloneable.hpp
@@ -42,7 +42,7 @@ namespace gstlrn
   };
 
 // std::remove_cvref_t not defined in C++17, however std::decay_t provides a
-// close-enough equivalent since this is not used here with functions or arrays, see:
+// close-enough equivalent since this is not used here with functions or arrays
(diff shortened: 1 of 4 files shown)


pierre-guillou (approved):
Thanks Rémi!

PR #930, opened 2026-06-02
Reduce memory usage for SPDE

remi-moyen:
A lot of classes uses internal work arrays to store intermediate values (for example when computing `out += A*B*in`, an intermediate array is needed to compute `B*in` before adding to `out`). When working on large data they can start to add up... a lot!

This PR tries to reduce memory usage in two ways:
- In some cases, some work arrays were allocated but not used (usually when they are needed in another function of the same class). Sometimes two different work arrays were used in two different functions when the same one could be reused. In some other cases a minor rearranging of the code can avoid using work arrays entirely.
- In two cases (`ProjComposition` and `ClassicalPolynomial`) I have implemented a pair of methods that allow the caller to specify the work arrays to be used, rather than allocating them locally within the class (this is entirely optional). When the caller knows exactly how these class are going to be used, and thus when each work array is going to be used, this allows sharing the same work array(s) between several instances of these classes, or reusing the work arrays for something else.

Changes related to the first point should not be too contentious. For the second point, this is slightly contorted and exposes some of the internals of classes in a way that isn't very nice (and puts some responsibility on the caller to ensure the work arrays are used properly!), but I think it works. It is entirely optional so I think it should be OK (just ignore those function when memory usage is not an issue).

### DIFF
--- a/src/LinearOp/PrecisionOpMulti.cpp
+++ b/src/LinearOp/PrecisionOpMulti.cpp
@@ -439,7 +439,7 @@ namespace gstlrn
     {
       messerr(
         "computeLogDet is not implemented for multivariate case in the "
-        "matrix-free version\n");
+        "matrix-free version");
       return TEST;
     }
     double result = 0.;
@@ -454,6 +454,9 @@ namespace gstlrn
   {
     if (idx < 0 || idx > _getNCov())
     {
+      messerr(
+        "The index passed to getPrecision() must be between 0 and %d\n",
+        _getNCov()-1);
       return nullptr;
     }
     return _pops[idx];
--- a/src/LinearOp/PrecisionOpMulti.cpp
+++ b/src/LinearOp/PrecisionOpMulti.cpp
@@ -450,12 +450,13 @@ namespace gstlrn
     return result;
   }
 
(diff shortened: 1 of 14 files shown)


src/LinearOp/PrecisionOpMulti.cpp
```diff
@@ -449,4 +449,14 @@ namespace gstlrn
...
   }
+
+  PrecisionOp* PrecisionOpMulti::getPrecision(Id idx)
+  {
+    if (idx < 0 || idx > _getNCov())
+    {
+      return nullptr;
+    }
+    return _pops[idx];
+  }
```
NDesassis, line 463: Could you add a warning  message (with function messerr) indicating that idx should be between 0 and _getNCov())?

NDesassis (changes requested):
Thx a lot Remi. Just a small warning message to add.

remi-moyen:
Thanks Nicolas. I've added the message.

I've changed another message to remove the trailing `\n`, I believe that since it is added by `messerr()` it should not be in the call, but maybe there was a good reason to have another one? (for pretty formatting of the message)

fabien-ors (approved):
Thx a lot Rémi

NDesassis (approved):

PR #937, opened 2026-06-15
Comparing different Krigings

drenard77810:
This branch is dedicated to a bug detected when comparing all kriging solutions:
- traditonal using Moving or Unique Neighborhoods
- SPDE based
- Vecchia based.

Several flaws are visible:
- Vecchia had to be corrected as it did not take the constant mean into account correctly. Can we envisage considering stringer drift hypotheses (OK, UK, KED)?
- Vecchia does not know abount the calculation of Variance of Estimation error, nor Variance of the estimator
- Vecchia should be robustified: when the number of Vecchia neighbors is larger than the number of data in dbin.


fabien-ors (approved):
Thanks Didier !!

PR #938, opened 2026-06-17
GitHub Actions: Use chocolatey to install Doxygen on Windows

pierre-guillou:
This PR switches to using the `choco` package manager on Windows to install a recent version of Doxygen rather than downloading a binary archive from `www.doxygen.nl` (which seems to implement some kind of rate limiting).

Corresponding `publish` workflow runs:
- https://github.com/pierre-guillou/gstlearn/actions/runs/27676708224
- https://github.com/pierre-guillou/gstlearn/actions/runs/27676702981

Pierre


fabien-ors (approved):
Thx Pierre

PR #941, opened 2026-06-18
Various fixes when HDF5 is disabled

remi-moyen:
Several minor changes when HDF5 is not available. A couple of tests still tried to call HDF5 functions and crashed.

Also the ASCII deserialization of `Vario` was wrong compared to the current ASCII serialization, I have tried to fix it.

But another issue was that the `vario.ascii` file in the test data does not actually matches the current ASCII serialization code, so I also updated it. Hopefully that doesn't break anything else...

### DIFF
--- a/doc/data/Scotland/vario.ascii
+++ b/doc/data/Scotland/vario.ascii
@@ -4,6 +4,9 @@ Vario
 1 # Number of directions
 0 # Scale
 1 # Calculation Flag
+# Variable names
+Unknown
+#
 # Variance
 0.363298 
 # Direction characteristics
--- a/include/Basic/ASerializable.hpp
+++ b/include/Basic/ASerializable.hpp
@@ -123,7 +123,11 @@ namespace gstlrn
 
   private:
     static String _myPrefixName;
+#ifdef HDF5
     EFormatNF _defaultFormatNF{EFormatNF::H5};
+#else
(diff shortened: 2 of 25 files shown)


remi-moyen:
I'm also using this opportunity to mention that with the current code, both on Linux and Windows, a few tests don't pass for me:

176 - test_LikelihoodFit_cmp (Failed)
190 - test_ModelFitting_cmp (Failed)
200 - test_OptimLikelihood_cmp (Failed)
202 - test_OptimVMap_cmp (Failed)

The output files show some minor numerical differences, I don't know to what they could be linked. Since tests seem to pass on here and for other devs (I assume?), and I don't think I have any local changes that could affect these, I'm not sure what the reason is. I'm assuming something in the numerical libraries (and therefore totally unrelated to this PR!) but I have no idea, really.

remi-moyen:
All "var" tests (*) were broken because of the fix in deserialization of `Vario`. Now it expects a `Variable names` section with one line per variable.

(*) except Var1, because it has 0 as "Calculation Flag" and because in the old code the variable names were only read when this flag was 1, I left that in the new code even though the serialization actually always write the variable names. So I think this test is also broken, or rather the deserialization code is still not quite correct and making it correct would break this test as well?

fabien-ors (approved):
Thx Remi.

PR #942, opened 2026-06-18
Fix compilation error on Windows/VS2017.

remi-moyen:
Not sure whether the issue is Windows or VS2017.

We had the same issue a long time ago and weirdly in `FileLAS.cpp` we still manage to use `std::tolower()` so this might just be a missing `#include`? In any case, the alternative (`::tolower()`) is already used in a few places so that should be fine.


fabien-ors (approved):
Thx Remi

Issue #939, opened 2026-06-18
Fix Vecchia if ndat < nb_vecchia


drenard77810:
Robustify: if ndat (number of Data in dbin) is lower than 'nb_vecchia'VarZ)

drenard77810:
This point has already been fixed. In fact, this was not an issue... but a consequence of counting the number of active samples in Dbout (by reading a Z-variable). If this variable was there in the output file before Vecchia Kriging, the result was not correct. This has been robustified already in another PR

PR #944, opened 2026-06-20
Correcting vecchia limitation of nb vecchia vs nech

drenard77810:
This branch is dedicated to a bug that was reported by Xavier when practicing Kriging using different techniques:
- traditional one: using Moving or Unique Neighborhood
- SPDE
- using Vecchia approximation

It appeared that Vecchia was sensitive to the number of points in its neighborhood (nb_vecchia) in particular if it was smaller than the number of conditioning data or not.

This remark was finally a misleading property:
The actual bug was linked to the sequence in which the Kriging were performed. When done after a serie of other Kriging, the output file did have some Z-variables. When entering in Vecchia a Ball tree is established based on the union of the input Db (optional) and the output Db (compulsory). The module for counting the number of samples (same for 'dbin' and 'dbout') was actually reading the Z-variable in order to discard potential TEST values. A special treatment was proposed if no Z-variable was found in the  file. This is relevant for the input Db, not for the output Db.

This has now been fixed.

The issue #939 is not closed voluntarily as it incorporates some other features that should be corrected or implmeneted before the issue can be fixed.


doc/demo/python/Oise_Final.ipynb
```diff
@@ -286,7 +286,8 @@
         "yy = np.linspace(6875000, 6982200, 2144)\n",
         "xx, yy = np.meshgrid(xx, yy)\n",
         "\n",
-        "angles.shape = (np.size(angles),)\n",
+        "angles = angles.ravel()\n",
+        "# angles.shape = (np.size(angles),)\n",
```
fabien-ors: Commented line to be removed

Could you squash the two last commits plz?

fabien-ors (changes requested):

fabien-ors (approved):
Thx Didier

PR #945, opened 2026-06-21
Marimo for PGS

drenard77810:
This branch is meant to develop a tool visualizing the PGS possibilities using MARIMO


fabien-ors (changes requested):
Could you squash undesirable commits having message like "Modif" etc.. plz ?

fabien-ors (approved):
Thx Didier

PR #947, opened 2026-06-23
Fix typo in ParamIdEqual comparison operator.

remi-moyen:
This is the root cause for #933 and is obviously wrong.


pierre-guillou (approved):
Thanks Rémi!

PR #948, opened 2026-06-23
add checkout step to html checker workflow

maxenceb91:
Add missing actions/checkout@v6 step to clone the repository onto the runner.


fabien-ors (approved):
Nice Maxence

PR #949, opened 2026-06-24
Add file folder verification in the html checker

maxenceb91:
The python script can now isolate the folder path from the href to verify if it's the correct one.


fabien-ors (approved):
Thx Maxence

Issue #871, opened 2026-02-18
Variogram cloud improvement


drenard77810:
These improvements are in three folds:
0) make this class ASerializable in order to allow storing the calculation parameters as well as the results (calculated on a grid). This would speed up the two next functions.
1) in the function db_vcloud(), allow a short cut where the calculation should be performed in a omni-directionnal manner (such as what has been done for Variogram calculations). The only parameter left is the maximum distance up to which pairs should be considered. 

2) add a function for Inquiry on the variogram cloud calculations.
This function should enable:
- creating a new Grid (containing the variogram cloud results) where the statistics are based on a limited set of selected samples: this would enable checking the impact of these samples on the variogram cloud
- asking for statistics of observation of samples in the pairs selected in a given area, selected by the user (min-max on distance and on variability): this would allow pointing out outliers for example.

These two capabilities would be an introduction to using the variogram cloud as an interactive Exploratory Data Analysis tool.

PR #951, opened 2026-06-24
Improving variogram cloud

drenard77810:
This branch partly answers to the Issue #871.

1) The function db_vcloud ()is now renamed vcloudFromDb().
The order of its arguments has been modified to allow 'lagmax' is higher order
The argument 'varioparam' is now optional: when not present, the omnidirectional 'varioparam' is created internally

2) The method vcloud_calculate() is now called vcloudFromParam()

3) The method vcloudGrid() (which was used only locally to dimension the output grid) has been deleted.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,8 @@ This file contains the logs between consecutive releases
 
 Release 1.12.0
 ==============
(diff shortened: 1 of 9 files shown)


fabien-ors (approved):
Thanks Didier!

PR #953, opened 2026-06-25
Improve hash function for more robustness

remi-moyen:
boost::hash_combine() tries to ensure all bits of the hash are mixed, so that when this 64 bits hash is folded back into the number of buckets (to decide which one to use), all buckets are equally likely to be used. A less robust hash would put several keys in the same bucket, which works but makes the hash map less efficient.


fabien-ors (approved):
Thx Remi!

PR #955, opened 2026-06-25
Revisiting Bench option in Variogram calculation

### LINKED ISSUE #870: More capabilities in Bench of Variogram
When calculating directional Variograms, we can use the qualifier 'bench'. 
This allows refusing any pair such that the distance (in a 'given' direction) between the two samples of the pair il larger than twice the size of the 'bench'.
This is very interesting in order to concentrate on pairs of samples belonging to the same layer ... in a sedimentary environment.

This is probably the reason why the famous 'given direction' has been selected as being the vertical in a 3D case. By extension it is the Y direction in the 2D case. More generally, in a NDIM case, it is the NDIM-th direction (1 based convention).

This scheme could be enriched by simply adding the possibility to add a 'Direction' (a simple vector in the NDIM space) along which this distance is calculated. We could then face tilted sedimentary environment.

drenard77810:
This branch revisit the Bench option used for Variogram calculation.
Initially the bench criterion was the following:
- considering the vector VEC between the Target and an eligible Data
- considering the width of the bench option (BENCH)
- considering the space dimension NDIM
The Datum was rejected if VEC[NDIM-1] > BENCH.
Clearly the criterion was applied on the last coordinate (vertical in 3D).

The demand was to generalize this criterion by defining in addition a vector for considering the bench criterion: BENCHDIR. This is vector provided by the user: 
- if it is provided, it is normalized internally (to speed up further calculations)
- if is is not defined, it is generated as the unit vector with a 1 in its NDIM-1 component (to be compliant with the previous implementation of the bench option).

Now the algorithm becomes:
ABS(BENCHDIR * VEC) > BENCH
(where "*" stands for the scalar product.

So doing, I also discovered that two elements were not saved/restored in Neutral Files: CYLRAD and BENCH.
They have been added.  The READING has been enhanced in order to remain compatible with the old files (where the two arguments were missing).

Fix #870.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,8 @@ This file contains the logs between consecutive releases
 
 Release 1.12.0
 ==============
-
+- In Variogram calculation with Bench criterion, BenchDir has been added
+- Bench and Cylrad criteria are now saved in Neutral Files (H5 format only)
 
 Release 1.11.1
 ==============
--- a/include/Basic/SerializeHDF5.hpp
+++ b/include/Basic/SerializeHDF5.hpp
@@ -163,6 +163,24 @@ namespace gstlrn
       return file;
     }
 
+    inline bool existsH5Item(
+      const H5::H5Object& grp,
+      const String& name,
+      bool verbose = false)
+    {
+      const auto grp_name = grp.getObjName();
+
+      if (!grp.attrExists(name))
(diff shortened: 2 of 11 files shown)


fabien-ors (approved):
Thx Didier!

PR #956, opened 2026-06-25
Refactor ShiftOpStencil / PrecisionOp / PrecisionOpMulti

remi-moyen:
Various refactoring of `ShiftOpStencil` / `PrecisionOp` / `PrecisionOpMulti`. The main change is that `ShiftOpStencil` now stores the stencil itself in a new class (`ShiftStencil`) which is independent of the `AShiftOp` itself.

### DIFF
--- a/include/LinearOp/PrecisionOpMulti.hpp
+++ b/include/LinearOp/PrecisionOpMulti.hpp
@@ -62,7 +62,7 @@ namespace gstlrn
     double computeLogDet(Id nMC = 1) const override;
     std::pair<double, double> rangeEigenVal(Id ndiscr = 100) const override;
 
-    PrecisionOp* getPrecision(Id idx);
+    PrecisionOp* getPrecision(Id idx) const;
 
   protected:
 #ifndef SWIG
--- a/include/LinearOp/PrecisionOpMulti.hpp
+++ b/include/LinearOp/PrecisionOpMulti.hpp
@@ -62,6 +62,7 @@ namespace gstlrn
     double computeLogDet(Id nMC = 1) const override;
     std::pair<double, double> rangeEigenVal(Id ndiscr = 100) const override;
 
+    Id getNCov() const;
     PrecisionOp* getPrecision(Id idx) const;
 
   protected:
@@ -72,7 +73,6 @@ namespace gstlrn
 
     void buildQop(bool stencil = false);
     Id size(Id imesh) const;
-    Id _getNCov() const;
 
     Id _getCovInd(Id i) const { return _covList[i]; }
 
--- a/include/LinearOp/PrecisionOpMulti.hpp
+++ b/include/LinearOp/PrecisionOpMulti.hpp
@@ -46,6 +46,10 @@ namespace gstlrn
       const VectorMeshes& meshes = VectorMeshes(),
       bool stencil = false,
       bool buildOp = true);
+    PrecisionOpMulti(
+      std::vector<PrecisionOp*> pops,
+      Model* model,
+      const VectorMeshes& meshes);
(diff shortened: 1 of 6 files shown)


include/LinearOp/PrecisionOpMulti.hpp
```diff
@@ -112,6 +116,7 @@ namespace gstlrn
     VectorInt _nmeshList;
     bool _allStat;
     bool _ready;
+    bool _destroyPrecisionOp;
```
pierre-guillou: Mixing ownership seems quite unsafe to me. Can't we add a vector of `PrecisionOp` as member that can own the memory pointed by `_pops` in one case, and make `_pops` points to caller memory in the other case? Destructor should be trivial this way.

remi-moyen: I'm not sure I understand your suggestion, is it something like this:

```c++
class PrecisionOpMulti {
  std::vector<std::shared_ptr<PrecisionOp>> _storedPops;
  std::span<PrecisionOp*> _pops;
};

PrecisionOpMulti::PrecisionOpMulti(... /* meshes etc. */) {
  // create individual PrecisionOp, store them them _storedPops
  _pops = _storedPops;
}

PrecisionOpMulti::PrecisionOpMulti(std::vector<PrecisionOp*> existingPops) {
  // _storedPops remains empty
  _pops = existingPops;
}

// rest of the code uses _pops only, destructor doesn't need anything to delete _storedPops as needed
```
(though actually if using `shared_ptr` to have the destruction happen automatically, the `span` needs to be on `shared_ptr` as well which requires the caller to use that as well)

This would work but I'm not sure it is much simpler than the `_destroyPrecisionOp` flag? One issue is that it requires the caller to necessarily keep the `vector` alive, and not just the individual `PrecisionOp` inside it (the caller may want to store them in a structure that isn't a `vector`).

I took the design with a flag from the way `PrecisionOp` stores the `ShiftOp`, it seemed more logical to do the same thing in several places?

pierre-guillou: Sorry for the late reply, was away last week.

That's indeed what I was proposing: removing the `bool` flag that is easy to misuse and rely on the type system to automatically release memory. With the caveat that the `std::span` should be a `std::vector` instead (still one allocation, but the pointers stored are shallow / non owning), thus decoupling from the external storage of `PrecisionOp` objects.

You can switch from `std::shared_ptr` to `std::unique_ptr` for `_storedPops` as I don't think you need the `std::shared_ptr` API, you just take references to stored objects. 

Similarly, `_pops` might also gained to be converted from `std::vector<PrecisionOp*>` to `std::vector<std::reference_wrapper<PrecisionOp>>`, which better highlights the non-owning nature of the pointers.

remi-moyen:
I have changed the code following your suggestion (I think). Hopefully this is now OK?

Two comments:
- I based my previous version on the design of `PrecisionOp` which has a flag `_destroyShiftOp`. Following your comments here, I think that `PrecisionOp` should also be changed in a similar manner.
- I did not rebase this branch to master (I'm not sure I can do a `git push --force-with-lease` and I'm not sure what updating in github will do exactly, rebase or merge), so I will let you do it if you accept the PR.

remi-moyen:
Sorry, I forgot to commit a couple of fixes. I thought they were already in the previous commit but just noticed now that they weren't!

pierre-guillou (approved):
Thanks Rémi!
I'll rebase your branch on top of `origin/dev` and will merge it once the CI has finished running (again...).

The suggested changes in `PrecisionOp` may wait for another PR. 

Pierre

PR #957, opened 2026-06-26
Krigingvecchia can now handle any Drift (in addition to constant Mean)

drenard77810:
This branch allows Kriging using Vecchia approximation, to handle a Model which can have:
- a constant mean when performing SK
- any drift when performing OK, UK or KED


fabien-ors (approved):
Nice Didier!

PR #958, opened 2026-06-27
Revisiting the variogram calculations

### LINKED ISSUE #871: Variogram cloud improvement

drenard77810:
In this branch:
- we regrouped all the API into a new file Variogram/Variograms (hpp and cpp),
- we renamed the variogram Cloud functions: vcloudXXX
- we renamed the Variorgam Map functions: vmapXXX
- we improved the Variogram cloud calculations by:
  . providing additional functionalities for selecting several pair and derive statistics on the most frequently involved samples. This selection can be defined through:
    . a polygon
    . focusing on pairs at a distance smaller than 'distmax' and a variability above 'varmin'
  . Allowing the partial calculation: between the whole data set on one side and a limited number of samples specified by the user on the other hand.
The specific Tuto_VCloud.ipynb has been developed (in python).

Fix #871


src/Variogram/AVario.cpp
fabien-ors, line 1: Why don't converting the functions in static as proposed by the linter?

fabien-ors (changes requested):

src/Variogram/AVario.cpp
drenard77810: Long discussion:
I tried initially ... as the 8 warning in this code anoided me a lot!
When I pass to:
    static double _evaluateScaleVariogram(double w1, double w2);
in the hpp, then I have 5 new errors this time, saying that:
Assigning to 'double (gstlrn::AVario::*)(double, double)' from incompatible type 'double (*)(double, double)'
pointing at line such 257/
        _evaluateScale = &AVario::_evaluateScaleVariogram;

Then I was not able to solve this issue... until I asked ChatGPT who told me that, some times, we should bypass the severity of the linter.
This was my concludion too.

If you find a better solution, take the lead on this piece of code. I think that my solution is very light an suppresses the warnings and seems to be limited to a single line:
NOLINTNEXTLINE

fabien-ors (approved):
Thx Didier

PR #961, opened 2026-06-29
Add support for a CMAKE_INSTALL_ROOT user-defined variable (if CMAKE_INSTALL_PREFIX is not defined), for automatically versioned installs.

remi-moyen:
If CMAKE_INSTALL_ROOT is defined, the install directory is automatically generated as e.g., gstlearn-1.2.3a1. This avoids having to change the CMAKE_INSTALL_PREFIX whenever the version changes. If CMAKE_INSTALL_PREFIX and CMAKE_INSTALL_ROOT are both defined, the CMAKE_INSTALL_PREFIX takes precedence as this is the standard CMake mechanism.


fabien-ors (approved):
Lgtm. Thx Remi

PR #964, opened 2026-07-07
Db second version

drenard77810:
This is the first push on this new branch.
Please review it as:
- it fixes the names (directory DataBase), files (DbData and DbCol), the names of the arguments: icol and iversion
- still to debate if we call the emethods: addArray or addColumn...
- the order of arguments (in setValue of setArray)
- the preliminary work on SWIG incorporation: we still may improve by using direct implementation within swig of std::optional...

This means that it is now or never that we choose these names before they are spread all over!

In next steps:
- I will incorporate the new class Matrix2D of @pierre-guillou to take the inversion into account properly (nothing is done for the moment, except passing args)
- incorporate the new class Categorical when @pierre-guillou  is ready
- Add the decorative features that we discussed: NAAllowed, ...
- Handle the Locators ... probably renamed as Roles at that time
- Incorporate DbData (in a smooth manner as @pierre-guillou  did already in his branch) in order to switch from old to new structure (at least as far as Double values are concerned).


pierre-guillou (changes requested):
* What's a PO?
* I don't see my authorship information in the Git commits
* This is conflicting with my own work on `Db` integration, `Array2D` and Categories

drenard77810:
PO is a typo for PG (alias for Pierre Guillou): this solves your first two remarks Pierre!

I tried to incorporate your work which (for my understanding) is included in db_het.
I see no trace of Categories nor Array2D.
Am I missing something

fabien-ors:
Next steps for this PR:
- Squash all the commits
- Once the POC is ready, rebase your branch from [pierre-guillou/db_het](https://github.com/pierre-guillou/gstlearn/tree/db_het) branch in order to preserve original authorship.
- Fix conflicts to preserve your added features
The best is to wait the full POC availability before applying these steps in order to prevent you from reintegrating your code several times.

Note: Currently I don't understand why non-reg check_code and non-reg tests under windows MSVC are failing. I must investigate a little bit more.

PR #965, opened 2026-07-08
Documentation for R package

maxenceb91:
Adds a python script that generates R package documentation in the form of .Rd files in the package's “man” folder by parsing the XML generated by Doxygen.  

The script parses the XML, and when it finds something—whether it’s a class, an enumeration, or a function—it creates an object with the corresponding attributes found in the XML (if a function is associated with a class, it is linked to that class through its attributes). 

Added another python script that converts the R README.md file to .Rd format so it can be viewed in the browser using the `?gstlearn` command in the R console.


fabien-ors (approved):
Nice job Maxence!
Thank you very much for your work.

PR #966, opened 2026-07-10
Python documentation with sphinx

maxenceb91:
Improvements to the python documentation generated with Sphinx using a python script in the `python/doc` directory that automatically modifies the `__init__.py` file before the documentation is generated.

The script also modifies the Sphinx .rst source files to automatically generate a table of contents listing all the classes in the python package.


python/doc/source/conf.py
```diff
@@ -0,0 +1,238 @@
...
+def autodoc_skip_member_handler(app, what, name, obj, skip, options):
+    short_name = name.split(".")[-1]
+
+    if short_name == "thisown":
+        return True
+
+    if short_name.startswith("E_"):
+        return True
+
+    functions_to_skip = [
```
fabien-ors, line 42: Why these functions must be skipped?

fabien-ors (changes requested):
Just a small question please.
Thank you

python/doc/source/conf.py
maxenceb91: To avoid duplicating functions that are in groups

fabien-ors (approved):
Thanks a lot Maxence!

PR #969, opened 2026-07-16
Add Python dependency for doxy2rd

fabien-ors:
The new feature for generating R package documentation from doxygen output was broken because Python was not available for executing doxy2rd.py script.
CMake find_package instruction to look for a minimal 3.11 python has been added to fix this issue.


drenard77810 (approved):
Thanks

PR #970, opened 2026-07-17
New data base

drenard77810:
This branch is dedicated to the adaptation of the initial code from @pierre-guillou proposed in #968 concerning the New Heterogeneous DataBase. This PR 
 - add the concept of Role which will replace the Locator terminology (in a future PR)
 - add the RoleID class which offers a generic way to identify a Role
 - add to the DbData class :
    - the capability to designate the columns using a unique ColID (which can be built from a column index, its unique name or a specific RoleID)
    - the possibility to set or get a value in the database using C++ overridden operator ()
 - implements typemaps for ColID and RoleID
The test test_Db_het.cpp (initially designed by @pierre-guillou) has been deleted and incorporated in the larger test_DataBase.cpp. Equivalent tests for python and R have been added.

PR #972, opened 2026-07-29
Make marimo interactive

drenard77810:
Adding 2 Marimo scripts:

- The good-old Playkrige facility is now hosted in Mo-Playkrige.py
- A new script allows demonstrating the Potential interpolation method.

I also had to enlarge the capabilities of:

- Db in order to offer the possibility to change the number of Columns / Rows of an existing Db
- Polygon in order to correct the calculation of the Polygon centroid: this required for "strange" Polygons read to display coast lines on the world map. Some of the vertices have a large unpredictable value (to mention the break of a Polygon into several Polysets). So the centroid calculation (center of gravity of polygon vertices) must discard these strange coordinates.


pierre-guillou (changes requested):
- Changes in `doc/courses/python/` & `doc/demo/python` seem useless and unrelated to the current PR
- Should `ipynb` files store their output?
- Should `NF` files in `doc/UI/python` be part of this PR?
- Files `python/doc/source/classes.rst` and `python/modules/=` don't seem useful or in the scope of the PR
- Changes in `DbCol.hpp` seems useless and prone to conflict with #970 
- Changes in `Db`, `CalcKriging`, `Estimations`, `PolyElem` and `ShiftOpStencil` are not mentioned in the PR description
- Commit messages should be in English and properly describe the corresponding changes

drenard77810:
My comments to your comment (when relevant)

-    Changes in doc/courses/python/ & doc/demo/python seem useless and unrelated to the current PR.
-     Should ipynb files store their output?
They are useless. We can regret that, when running any ipynb file (in order to check its behavior) although not modifying any part of the script, the ipynb file is considered as "Changed" and registered as such (if the user does not pay attention to the list)

-     Should NF files in doc/UI/python be part of this PR?
They are created automatically (if the option is activated in the Marimo script). When running the Marimo script directly in the distribution directory of gstlearn, they are registered as "Changed" and registered as such (if the user does not pay attention to the list)
-     Files python/doc/source/classes.rst and python/modules/= don't seem useful or in the scope of the PR
Mistake. Could be interesting to detect which action creates them in order to correct the creation of these spurious file.
-     Changes in DbCol.hpp seems useless and prone to conflict with https://github.com/gstlearn/gstlearn/pull/970
I should not modify DbCol.hpp in this branch. This is reserved for the branch names "New_DataBase".
-     Changes in Db, CalcKriging, Estimations, PolyElem and ShiftOpStencil are not mentioned in the PR description
I added this information (in the first comment) of this PR
    Commit messages should be in English and properly describe the corresponding changes
    I plan to squash them out when the PR corrections are finished.

pierre-guillou (approved):

PR #976, opened 2026-08-18
GitHub Actions: Make CMake find Sphinx in publish_python_windows

pierre-guillou:
This PR sets the CMake variable `Sphinx_DIR` to make CMake find the installed `sphinx-cmake` files in the `uv` virtualenv.

In addition, the `Python3_ROOT_DIR` CMake define has been removed, since the environment variable `pythonLocation` is not set anymore and the double quotes prevent the variable expansion.

Corresponding `publish` run: https://github.com/pierre-guillou/gstlearn/actions/runs/32154253886/job/95767490083

Pierre


fabien-ors (approved):
Thks Pierre!

PR #978, opened 2026-08-24
Improving playkrig marimo

drenard77810:
This new branch concludes the development of Playkrige. 
It also served in reviewing all the previously established Marimo script in order to update them.

Important remark: it lead to one correction in gstlearn Db / DataBase facilities. This could be moved into another branch if needed. This correction is essential to avoid a severe crash (of krigWeights() method for example ... but not only).

It should be merged AFTER the acceptation of the main developement branch named New-DataBase.


drenard77810:
The essential modification of setValueByUID (in Db.cpp) has been duplicated in the branch New-DatBase (where it ought to be initially).

PR #982, opened 2026-08-31
Fix unused static variable warnings coming from MacOS compiler

drenard77810:
This branch is meant to check the new errors essentially due to compiler modifications.


fabien-ors (approved):
Thx Didier!

PR #983, opened 2026-09-02
Add VMap on a Flumy slice

fabien-ors:
This PR adds a new metric in the sandbox ipynb demo to show how to measure some properties on a flumy simulation


drenard77810 (approved):
Thanks

PR #984, opened 2026-09-02
Workflow optimization

drenard77810:
This branch is a first step towards optimization.
It consists: 
-developing three new scripts (in Python) meant to estimate and visualize the computing time of a set of workflows across time
- moving the typemaps (developed specifically to interface ColID and optional to Python and R) to the language-dedicated directories rather than in the Swig directory
- trying to improve the aforementioned typemaps in order to gain computing time.


fabien-ors (approved):
Thx Didier

PR #981, opened 2026-08-28
Sb data extension for categorical variables

drenard77810:
This is the last step for integrating the Categorical variables in the new DataBase.
The complete non-regression tests have been completed (they now incorporate special paragraphs for testing these Categorical Variables and the Dictionary).
The relevant update of the typemaps have been incorporated.

PR #985, opened 2026-09-02
Cleaning db by calling dbdata directly

drenard77810:
In this branch, we started regrouping the methods in Db which are direct link to the underneath DbData.
In a future version (when green light is given), these multiple methods will possibly be replaced by a single one (using ColID versatile argument).
In details, instead of:
- getValue(const String& name, ...)
- getValueByColIdx(Id icol, ...)
-getValueByUID(Id iuid, ...)
we will simply call:
- getValue(const ColID& colid, ...)

Additionally some duplicated functions have been found: one of them is placed as DEPRECATED (it will be deleted in a future version)


.github/workflows/nonreg-tests_macos-latest.yml
```diff
@@ -75,6 +75,9 @@ jobs:
           -DBUILD_R=ON \
           -DSWIG_EXECUTABLE=${{env.SWIG_ROOT}}/bin/swig
       env:
+        # Next line is meant to let ALL the warnings to be trapped before job stops
```
fabien-ors: to be removed.
Already in PR #981

src/Db/Db.cpp
```diff
@@ -6355,23 +5668,685 @@ namespace gstlrn
...
+
+  Id Db::_getColumnFromLocator(const ELoc& locatorType, Id locatorIndex) const
+  {
+    auto colID = _data.getColID({temporaryToRole(locatorType), locatorIndex});
+    auto icol = colID.getICol();
+    _debugConditionalStatement(icol);
+    return icol;
+  }
+
+  Id Db::_getColumnFromColID(Id icol0) const
```
fabien-ors: I prefer keeping icol for the argument name.
No need to redefined a new local variable icol.
The argument is already a copy.

src/Db/Db.cpp
```diff
@@ -6355,23 +5668,685 @@ namespace gstlrn
...
+    return icol;
+  }
+
+  String Db::getNameByLocator(const ELoc& locatorType, Id locatorIndex) const
+  {
+    auto icol = _getColumnFromLocator(locatorType, locatorIndex);
+    return _data.getName(icol);
+  }
+
+  String Db::getNameByColIdx(Id icol0) const
```
fabien-ors: idem etc.

include/Anamorphosis/AAnam.hpp
fabien-ors, line 1: irrelevant change

include/Db/Db.hpp
```diff
@@ -278,96 +278,174 @@ namespace gstlrn
...
 
+    void setLocator(
+      const String& name,
+      const ELoc& locatorType,
+      Id locatorIndex = 0,
+      bool cleanSameLocator = false);
     void setLocatorByUID(
       Id iuid,
-      const ELoc& locatorType = ELoc::fromKey("UNDEFINED"),
+      const ELoc& locatorType,
```
fabien-ors, line 321: Why removing the default value of  = ELoc::fromKey("UNDEFINED" ?
This implies a lot of modifications in the code (interface break)

src/Db/DbHelper.cpp
```diff
@@ -719,7 +719,7 @@ namespace gstlrn
     for (Id icol = 0; icol < ncol; icol++)
     {
       jcol = cols[icol];
-      if (!db->isColIdxValid(jcol))
+      if (!db->getData().isValidColumn(jcol))
```
fabien-ors: Internal member exposure (getData) should be avoided.
Prefer keeping the method isColIdxValid in the Db class (and directly call the one from DbData)

src/Stats/Classical.cpp
```diff
@@ -1614,7 +1614,7 @@ namespace gstlrn
     for (Id icol = 0; icol < ncol; icol++)
     {
       Id jcol = cols[icol];
-      if (!db->isColIdxValid(jcol))
+      if (!db->getData().isValidColumn(jcol))
```
fabien-ors: idem

tests/py/test_Kriging.py
```diff
@@ -210,11 +210,11 @@ def test_kriging_internal(
 
     # Creation of a db2 without selection to build the complete covariance matrix
     db2 = db.clone()
-    db2.setLocator("sel")
+    db2.setLocator("sel", gl.ELoc.UNDEFINED)
```
fabien-ors: Prefer calling resetLocator or deletelocator for that operation.
(everywhere)

fabien-ors (changes requested):
Thank you Didier
Some small changes requested please.

include/Db/Db.hpp
drenard77810: 1) It is done and modifying backwards is always dangerous.
2) Moreover I do not like (anymore) using a missing argument to specify that we want to clear the ELoc. 
3) If you do not like setting the LocatorType back to ELoc::UNDEFINED (which is actually what happens in the code) by saying it explicitly, I propose using a more straightforward method called clearLocator(const String& name). I already coded it and plugged it on the code.

fabien-ors: Agree

fabien-ors (approved):
Thanks Didier!

PR #986, opened 2026-09-07
Add nversion to addColumn

drenard77810:
This branch is meant to experiment the use of 'nversion' when using a variable in Db. It is performed in several steps:

1) Allow using 'nversion'
- The first stage was to add 'nversion' as an additional argument to addColumns() method (and the sister methods such as addColumnsByVVD(), addColumnsByConstant() and addColumnsRandom())
- as a consequence, the similar improvement has been done in protected methods _addVariableDb() of ACalcDbToDb and ACalcDbVarCreator classes.


drenard77810:
This is the end of the first phase (TEDIOUS).
In fact, I tried to keep the modified files as few as possible... and I only modified 180 of them!
These are the reasons and my comments:
0) In many methods, we had to add the "version" argument. This argument was placed as the last argument (with a default value set to 0) in order to keep the scripts unchanged as much as possible. This temporary patch will have to be deleted in the final version, where an item of the database will be refered to using the ColId argument (which incorporates the version) rather than the traditional name of locator or column rank (direct of thru UID). This means that the final modification will apply AGAIN on almost 200 files!
1) I had to modify the display of Db when statistics are required. As a matter of fact, in the olf version, all columns were considered. In the new version, I had to go in a loop over the Columns, and potentially over the versions of a Column.
Then, for each item, displaying the LOCATOR became irrelevant: a locator is assigned to a column, not to a version of a column. Then I decided to suppress all information concerning the locator from this display. But this display is involved in a large series of outputs of non-regression tests!
2) The good old operator [] which was active to address the contents of a variable has not been improved in order to address any version within any column. This improvement will have to be imagined in the final version: scripts had to be adapted in the mean time.
3) The most important adaptation were brought to the plot.py and plot.R facilities concerning the use of the version. This will be covered by itself, as soon as the ColID argument is used (instead of the traditional column "name").
4) A new limitation appears in the setName() facility where we cannot use this method to rename ONE simulation outcome in particular: the whole column is now renamed (i.e. all the simulation outcomes
5) The variogram calculation when applied to the outcomes of a simulation required all of the corresponding variables either to be specified by their name (or globally through a locator). Obviously, this is not valid anymore. So I has to create the (temporary,) method extractVersion() which duplicates one version of a Column as a new fresh variable. Of course, the best solution in the final version will be to perform the loop on the simulation outcomes internally to the variogram calculation. (As a simplification, we can probably admit that, in that case, all the versions will be systematically implied).
6) Now that the simulation index (considered as the 'version' index) is separated from the index of the GRF or the PGS (considered as the 'variable' name), we could spot an OLD bug in the calculations of PGS. This is fixed now (but implied a modification of the non-regression reference files)and MAF calculations

PR #987, opened 2026-09-08
GitHub Actions: Fix publish workflow

pierre-guillou:
This PR fixes two issues preventing the run of the `publish` workflow.

- The first one hit the `call_python_linux` job. Seemingly due to the SWIG 4.4 to 4.5 upgrade, lines were inserted in `__init__.py` at the wrong position, leading to an on-the-fly Python indentation error detected by Doxygen. The inserted lines are now in the correct position but since it is hard-coded, it remains very brittle.
- The second one hit both `call_macos` workflows (only the `x86_64` variants) and is caused by the Doxygen upgrade from 1.17 to 1.18. The offending code in Doxygen has been fixed in https://github.com/doxygen/doxygen/commit/26e908779f448574c6b8918aff21a40565a89a57 (`size_t` was not compared to `npos`) but we need to wait to the next Doxygen release or rollback to 1.17. The latter is done in this PR, with the help of `pixi` (`brew` doesn't provide Doxygen 1.17 anymore).

In addition, function names have been sorted alphabetically in python/doc/source/conf.py.

Corresponding `publish` workflow: https://github.com/pierre-guillou/gstlearn/actions/runs/34249126808

Pierre


fabien-ors (approved):
Very Nice Pierre and many thanks !

